PR #1, opened 2023-12-08
logical operators for random select filters

MatVeiQaaa:
A feature to add logical operators to random select filters, particularly for keys with integer values. Implements "<", ">", "<=", ">=" and "&&" operations. 

Normally, beatoraja as of 0.8.6 allows for fixed value keys in random select filters, such as "clear": 6, which would select a song with Hard-Clear lamp. It isn't possible to have it select from a range of values, such as 4-5, which would describe all songs with lamp ranging from Easy-Clear to Groove-Clear. This PR allows doing it as such:
```json
{
        "name":"RANDOM EASY-GROOVE",
        "filter": {
            "clear": ">=4 && <6"
        }
 }
 ```
 This has no effect on normal beatoraja behavior, otherwise. Everything that would work in upstream still holds, it only expands on it.

### DIFF
--- a/core/src/bms/player/beatoraja/select/BarRenderer.java
+++ b/core/src/bms/player/beatoraja/select/BarRenderer.java
@@ -414,7 +414,7 @@ public class BarRenderer {
 				String getterMethodName = "get" + key.substring(0, 1).toUpperCase()
 						+ key.substring(1);
 				try {
-					if (this.getFilter().get(key) instanceof Integer) {
+					if (this.getFilter().get(key) instanceof Integer) { // Fork for integer values of the filter key.
 						Integer value = (Integer)this.getFilter().get(key);
 						if (scoreData == null) {
 							if (0 != value) {
@@ -430,15 +430,13 @@ public class BarRenderer {
 						return true;
 					}
 					Object valueArr[] = ((String)this.getFilter().get(key)).split("&&");
-					for (Object value : valueArr)
+					for (Object value : valueArr) // Fork for string values of the filter key.
 					{
-						value = ((String)value).replaceAll("\\s","");
+						value = ((String)value).replaceAll("\\s",""); // Clean from whitespaces.
 						if (scoreData == null) {
-							if (value instanceof String) {
-								String stringValue = (String) value;
-								if (!stringValue.isEmpty() && !stringValue.substring(0, 1).equals("<")) {
-									return false;
-								}
+							String stringValue = (String) value;
+							if (!stringValue.isEmpty() && !stringValue.substring(0, 1).equals("<")) {
+								return false; // Because lack of value would be less than anything.
 							}
 						} else {
 							Method getterMethod = ScoreData.class.getMethod(getterMethodName);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: core/src/bms/player/beatoraja/select/BarRenderer.java [hunk 1]
```diff
@@ -407,6 +407,93 @@ public Map<String, Object> getFilter() {
...
+							}
+						}
+						return true;
+					}
+					Object valueArr[] = ((String)this.getFilter().get(key)).split("&&");
+					for (Object value : valueArr)
+					{
+						value = ((String)value).replaceAll("\\s","");
+						if (scoreData == null) {
+							if (value instanceof String) {
```
seraxis, line 437: As you cast value to a String on line 435 this if statement will always return True, were you intending to catch Null's here? Not sure I understand

core/src/bms/player/beatoraja/select/BarRenderer.java
```diff
@@ -407,6 +407,93 @@ public Map<String, Object> getFilter() {
...
+							}
+						} else {
+							Method getterMethod = ScoreData.class.getMethod(getterMethodName);
+							Object propertyValue = getterMethod.invoke(scoreData);
+							if (propertyValue instanceof Integer) {
+								String valueString = (String)value;
+								Integer propertyValueInt = (Integer)propertyValue;
+								Integer filterValueInt;
+								if (valueString.substring(1,2).equals("=")){
+									filterValueInt = Integer.parseInt(valueString.substring(2,valueString.length()));
```
seraxis, line 451: You don't need to include the length parameter if you are taking a substring to the end of the string

core/src/bms/player/beatoraja/select/BarRenderer.java
```diff
@@ -407,6 +407,93 @@ public Map<String, Object> getFilter() {
...
+								}
+								else filterValueInt = Integer.parseInt(valueString.substring(1,valueString.length()));
```
seraxis, line 453: same as above

core/src/bms/player/beatoraja/select/BarRenderer.java
```diff
@@ -407,6 +407,93 @@ public Map<String, Object> getFilter() {
...
+		public Boolean filterSong(ScoreData scoreData) {
+			Set<String> filterKey = this.getFilter().keySet();
+			for (String key : filterKey) {
+				String getterMethodName = "get" + key.substring(0, 1).toUpperCase()
+						+ key.substring(1);
+				try {
+					if (this.getFilter().get(key) instanceof Integer) {
+						Integer value = (Integer)this.getFilter().get(key);
+						if (scoreData == null) {
+							if (0 != value) {
```
seraxis, line 420: instead of doing this if you can `return 0 != value` directoly

seraxis (changes requested):
Left some comments on the code, otherwise looks good to merge

core/src/bms/player/beatoraja/select/BarRenderer.java
MatVeiQaaa: That is a left-over from other approach, thanks for noting it.

PR #2, opened 2023-12-10
fix hispeedfix display without selected song

MatVeiQaaa:
Normally, Hi-Speed Fix mode isn't displayed in song select menu, if current bar doesn't belong to a song or a course. However, it is still possible to set this value by pressing START+5 - Start menu will not display current value in that case, but it will still update it for 7K mode. 
Every other option is displayed regardless of current bar type and defaults to displaying the value for 7K, if the option is keymode specific.
This commit changes the behavior for hispeedfix getter, so it matches that of other options.


seraxis:
Looks good, thanks

PR #4, opened 2023-12-21
Useful error message on failed bms parse

hadronyche:
Catch NPE on failed bms parse. Very rare but happens with some `bmson` charts, also added to `bm*` for posterity.

Will log the offending charts rather than an ambiguous NPE trace some function calls higher.


seraxis:
Looks good to me

PR #5, opened 2023-12-21
Add noGapX to remove width gaps in note graphs

hadronyche:
Remove measure gaps for cleaner notegraphs. These typically appear heavily aliased in skins.

Before:
image
After:
image

Measures can still be distinguished by heights of different bars and colour within.

### DIFF
--- a/core/src/bms/player/beatoraja/play/PracticeConfiguration.java
+++ b/core/src/bms/player/beatoraja/play/PracticeConfiguration.java
@@ -44,9 +44,9 @@ public class PracticeConfiguration {
 	private PracticeProperty property = new PracticeProperty();
 
 	private SkinNoteDistributionGraph[] graph = { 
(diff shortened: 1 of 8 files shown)


seraxis:
This definitely looks a lot cleaner, I'm going to experiment a bit with this change first before merging

hadronyche:
noGap is already a skin parameter, I've just changed it to getting rid of all the gaps instead of just in the y-direction. I was thinking of isolating this change to just the width (additional noGapX?) to preserve the original behaviour but figured this would be simpler.

For reference, the vertical gaps are these, which also alias:
image

seraxis:
A noGapX skin property would be in keeping with current behaviour and wouldn't surprise someone switching to Endless Dream

seraxis:
Does the state of the this PR match what was [merged in upstream](https://github.com/exch-bms2/beatoraja/pull/781)?

hadronyche:
Yeah, it does. Might be convenient to merge it now instead of waiting for the next raja.

seraxis:
Merging this in anticipation of upstream 0.8.7

PR #6, opened 2023-12-22
Global FREQ+

hadronyche:
Added a global set-and-forget FREQ+, integrated into the random trainer:
image

Additional considerations:
- Limit the slider to steps?
    - Currently, any integer between 100-200
    - Sliders can be Ctrl-clicked for manual input
- Skinnable FREQ+ %
- More testing
    - During dev it messed with the practice mode freq, but I think it's isolated now.
- Clean up code

### DIFF
--- a/core/src/bms/player/beatoraja/ImGuiRenderer.java
+++ b/core/src/bms/player/beatoraja/ImGuiRenderer.java
@@ -13,6 +13,7 @@ import com.badlogic.gdx.controllers.ControllerManager;
 import com.badlogic.gdx.controllers.Controllers;
 import com.badlogic.gdx.controllers.desktop.JamepadControllerManager;
 
+import com.sun.javafx.util.Utils;
 import imgui.*;
 import imgui.flag.*;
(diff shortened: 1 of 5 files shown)


hadronyche:
Doesn't work, chart render and actual chart are disconnected. Will investigate later.

hadronyche:
It's in a good spot now, worth testing.

seraxis:
Wow, great work, I'd like to integrate this into a refactor of the ImGui menu into submenus so this will remain open for the time being. Might be best if I opened a tracking PR with these changes integrated into the refactor?

hadronyche:
Agree, I had the same thought about sub-menus or something like collapsible sections for different features. Would be good to generalise the trainer menu a little.

hadronyche:
Testing right now, there's an issue where it doesn't apply ran. Probably because I've overwritten whatever PatternMod applied. There's also one chart I've found that fast forwards to 20% through instantly. Working otherwise.

seraxis:
closed, superseded by #9

PR #16, opened 2024-02-19
LR2 judge autoadjust logic

MatVeiQaaa:
I studied autoadjust logic in LR2 disassembly and recreated that logic here.

It fixes an issue of beatoraja autoadjust being too reactive, which results in too big changes of adjust at bursts, which results in a net negative for score.
On the other hand, it won't be able to reach the generally correct offset as quickly, but it's not as important. Previous logic is very capable of completely ruining the plays, which is the reason to choose against it for many.

Doesn't impact the game in any other way. Currently is under user testing.

Reference disassembly with comments in the attached screenshot.
image

### DIFF
--- a/core/src/bms/player/beatoraja/play/JudgeManager.java
+++ b/core/src/bms/player/beatoraja/play/JudgeManager.java
@@ -150,6 +150,8 @@ public class JudgeManager {
 	 */
 	private int recentJudgesIndex = 0;
 
+	private int pressesSinceLastAutoadjust = 0;
+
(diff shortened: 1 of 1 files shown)


MatVeiQaaa:
Maybe it makes sense to cut-off the early adjust in the range of 1-3/1-4? One obvious issue with LR2 autoadjust is that players generally hit a few milliseconds early, say a standard deviation. This results in auto-adjust setting the adjust too much into the negative and you get quite a bit more slows than fasts because of it. It's misleading, as normal human being sees more slows than fasts and thinks "i should hit earlier", which in reality only makes the problem worse.

PR #17, opened 2024-02-21
IR sendPlayData use flow rework

MatVeiQaaa:
Original beatoraja code sends the scores only on prepare() method of MusicResult scene. Before the send thread finishes its job, result screen will never let you go back to music select, so in the case IR lags, you have to wait until timeout. Failed to send scores are then resent on next MusicResult.prepare() call.

This PR allows to return to song select if the score takes more than 3 seconds to submit.
Failed to submit scores are resent to IR every 4^retry seconds in a background thread, which starts if connection to IR on game start succeeded. 
Code to resend scores from result scene is deleted, only the immediate score is sent from result screen. 
The number of maximum retries is set in "config_sys.json" file in the parameter "irSendCount", default value is 5. After failing to send the score for 5 times total, it's gonna be removed from the array, which is a behavior identical to original beatoraja flow. To prevent scores from being removed too soon in case of short IR or user connection downtime, the scores are sent in intervals of 4^retry seconds. With default irSendCount value of 5, it would take 1364 seconds to drop the score, or roughly 23 minutes.

### DIFF
--- a/core/src/bms/player/beatoraja/MainController.java
+++ b/core/src/bms/player/beatoraja/MainController.java
@@ -42,6 +42,8 @@ import bms.player.beatoraja.song.*;
 import bms.player.beatoraja.stream.StreamController;
 import bms.tool.mdprocessor.MusicDownloadProcessor;
 
+import static bms.player.beatoraja.skin.SkinProperty.*;
+
 /**
  * アプリケーションのルートクラス
  *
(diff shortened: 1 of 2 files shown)


core/src/bms/player/beatoraja/MainController.java
```diff
@@ -42,6 +42,8 @@
 import bms.player.beatoraja.stream.StreamController;
 import bms.tool.mdprocessor.MusicDownloadProcessor;
 
+import static bms.player.beatoraja.skin.SkinProperty.*;
```
MatVeiQaaa: :facepalm:

MatVeiQaaa:
Should it allow to leave result screen instantly, instead of waiting for 3 seconds? I made the 3 second wait to give it a chance to send the score and update the ranking table in that time, i deemed it to be the maximum acceptable wait. Maybe it's actually unnecessary?

seraxis:
Fantastic PR, I understand there are still some concerns with updated scoreboards but I think this PR is a good basis to do that work on. Merged

PR #18, opened 2024-02-22
IR send rework: bugfix for multiple IRs

MatVeiQaaa:
I made an oversight in the PR that is already merged, it had a bug that it would never send a score for any but the primary IR server, in case multiple IRs are connected.


seraxis (approved):
Looks good, nice catch

PR #19, opened 2024-02-23
FREQ score save bug

MatVeiQaaa:
This PR fixes a bug where a score wouldn't be saved when you have negative FREQ value set in rate modifier menu and rate disabled.


seraxis:
Nice catch! Merged

PR #22, opened 2024-03-18
Add IntelliJ run configurations for building with Gradle

corndogit:
This PR removes the catch-all exclusion of the .idea folder while following the [gitignore.io](https://www.toptal.com/developers/gitignore/api/intellij+iml) template for IntelliJ projects. This allows run configurations to be shared in repo, both for new contributors or for the sake of convenience.

## Changes

- Remove ignore rule for `.idea` as well as some duplicated rules in the main .gitignore file
- Add Gradle run configurations for Windows and Linux platforms based on [README guidance](https://github.com/seraxis/lr2oraja-endlessdream#building-from-source)
- Include vcs.xml (not completely necessary but it's not a high churn or user-generated file either)


corndogit:
Should also mention that this came off the back of looking for a way to include some project structure information in the repo, specifically defining source folders for the submodule dependencies. 

In summary it seems like there's no need to include the module dependency stuff in .idea if its already managed by Gradle, if anyone else runs into IDEA having issues about invalid `bms.model.*` or `bms.table.*` imports, maybe the step to mark the submodule `src` folders as sources could be added to the README.

Issue #20, opened 2024-02-27
Plays with Rate Modifier on always considered as a No Play Clear on Result screen


marysiamzawka:
![obraz](https://github.com/seraxis/lr2oraja-endlessdream/assets/20336327/5aa28a1e-be7d-41d9-bd16-7a37fbb4ee2d)
The issue occurs if you have Rate Modifier enabled and fail the chart (either organically or by quitting out early).
This could be fixed by applying Course-like behaviour to Rate Modifier fails (still is a No Play, but is registered as a Fail)

MatVeiQaaa:
Hmm, i agree it's an issue. Can implement a fix if it's decided to be unintended behavior.

seraxis:
This is intended behaviour, negative rate scores are never saved, in order for a fail lamp to be acquired for a song a score would need to be saved.

A potential compromise would be saving a fake zero'd out score with a fail lamp (so you could keep track of songs you've played on negative rates) but to me this doesn't make much sense.

Leaving this issue open for further discussion

MatVeiQaaa:
I think it should be possible to display a fail lamp without saving a score, or any lamp for that matter. I didn't test, but as far as o understand the piece of code for all that, you can store whatever data in temporary score, and then just not commit it.

seraxis:
If that's what Mary had in mind I wouldn't be against changing the current behaviour

MatVeiQaaa:
I'm gonna make a pr for that in couple of days then, if it works the way I think it does

MatVeiQaaa:
fixed with PR #24

Issue #8, opened 2023-12-24
Proposal: JUDGE+


hadronyche:
A trainer setting to target a given judgerank. The target will be applied if it's harder than the loaded chart's judgerank.

E.g. if I want to practice normal judge and set the target to normal judgerank: all easy judge charts will load as normal judge.

**(1)** Beatoraja currently lets you proxy this via `EXPAND JUDGE` and does not class it as an assist if all values are <=100. However, while it's simple to toggle on and off in-game, it's tedious to calculate the numbers (take from iidx.org) and input the numbers, and have to go into the launcher when you want to change them. A simple drop-down box to select judgeranks will be much more convenient.

Careful consideration should be given to how and whether this should affect charthash. Changing the judgerank in a `bms` file does affect the charthash. However, since the behaviour described in **(1)** is allowed without any repercussions (i.e. `ASSIST`, charthash change), there is a case to be made for it to not **(A)**. On the other hand, changing the chart hash will allow a separate tracking of scores on that judgerank **(B)**. Scores on higher judgeranks are difficult to view outside of result, but can be shown with a message on chart load (e.g. `EASY CLEAR, 4111 EX, 91.02% AAA`)

I've not investigated how possible behaviours **(A)** or **(B)** are to implement, but I would appreciate thoughts on which you think would be better to build for. My guess is that it's simple to implement and changing charthash dynamically should naively store the score in the DB, though there may be a conflict wherein a chart path with multiple charthashes may have undesired behaviour.

hadronyche:
Agreed, saving the local score as NO PLAY seems like the most simple, elegant solution and leaves extensibility open down the line.

An additional consideration is that most IRs support manual submission (LR2oraja: Tachi, CinnamonIR), which would send scores previously limited to local DB saves. This currently affects the Rate Modifier (to my knowledge, untested), which raises a discussion in itself about how much to cater to IRs and how this would limit future features. The use of FREQ in LR2 and tighter judges by EXPAND JUDGE in raja are obvious and allowed to submit on their IRs, but novel features without any precedent or parallel will require discussion.

I haven't investigated this but applying something like LIGHT ASSIST may allow IRs to ignore these scores. Otherwise, your idea of a separate DB would avoid this issue or we could bring features in line with what IRs accept to send scores.

Catizard:
[...]

PR #24, opened 2024-03-19
Issue #20: Plays with Rate Modifier on always considered as a No Play Clear on Result screen

MatVeiQaaa:
This makes the game display the fail lamp with rates enabled. All the other lamps will still be overridden into NO PLAY. In case of negative FREQ the lamp is not gonna be saved. In case of positive FREQ, the fail lamp will be saved together with the score, which i think is in line with original behavior for that scenario.

### DIFF
--- a/core/src/bms/player/beatoraja/result/MusicResult.java
+++ b/core/src/bms/player/beatoraja/result/MusicResult.java
@@ -438,7 +438,7 @@ public class MusicResult extends AbstractResult {
 			newscore = cscore;
 		}
 
-		if (isFreqTrainerEnabled()) {
+		if (isFreqTrainerEnabled() && resource.getScoreData().getClear() != ClearType.Failed.id) {
 			resource.getScoreData().setClear(NoPlay.id);
 		}
 


MatVeiQaaa:
honestly, why do we even save FREQ scores at all? Does it make any sense to save FREQ+ scores with NO PLAY lamp? It could be made so you don't get your scores saved ever if FREQ is enabled, but all lamps would be displayed on result screen normally.

hadronyche:
I think discussions on FREQ behaviour should be relegated to parity with LR2, since we're working on a build of LR2oraja, which aims to emulate LR2 behaviour. LR2 saves scores, lamps and submits to IR when using FREQ+. It also displays lamps on result as normal.

On FREQ- LR2 does not save scores and lamps, but it displays the score on result. The lamp, however, has some nuance, it only seems to display lamps when scores can be saved (see ops 62-66, tested on Seraphic result) and other lamp ops are reliant on music select (ops 100-105, from [here](http://right-stick.sub.jp/lr2skinhelp.html)) which enables display of best clear and current clear simultaneously. It may be possible to display lamp with ops 118-125 but I don't have time to test, I suspect these are used for gauge select.

MatVeiQaaa:
> I think discussions on FREQ behaviour should be relegated to parity with LR2, since we're working on a build of LR2oraja, which aims to emulate LR2 behaviour. LR2 saves scores, lamps and submits to IR when using FREQ+. It also displays lamps on result as normal.
> 
> On FREQ- LR2 does not save scores and lamps, but it displays the score on result. The lamp, however, has some nuance, it only seems to display lamps when scores can be saved (see ops 62-66, tested on Seraphic result) and other lamp ops are reliant on music select (ops 100-105, from [here](http://right-stick.sub.jp/lr2skinhelp.html)) which enables display of best clear and current clear simultaneously. It may be possible to display lamp with ops 118-125 but I don't have time to test, I suspect these are used for gauge select.

While generally i agree with that logic, i find it unnecessary to copy every bit of LR2 behavior, especially when it doesn't have much to do with gameplay itself. LR2 is full of bad decisions and behaviors inherited from code structure, copying them wouldn't be wise. Not to say this particular behavior is bad, but more of a call to also think outside this box.

hadronyche:
Agreed, LR2 shouldn't be copied 1:1 at face value and without careful consideration. However, I think the clear/score saving logic is intuitive and sound. FREQ- scores shouldn't be saved nor tracked because the modifier is a 'crutch' or practice tool (easier than the chart intends), whereas FREQ+ is almost always equal or more difficult than no FREQ modifier.

If we're considering this from a IR perspective, though I'm biased to being against catering to IRs, the two IRs, to my knowledge, that support LR2oraja (Bokutachi and CinnamonIR) both also support LR2 - which, again, records FREQ+ scores. As a thought experiment, if we were to grant that FREQ+ somehow is easier than FREQ 0 (say, player more comfortable at higher BPM) then those whose client uploads to IR would be at an advantage.

Regarding the result screen: Including as much information as possible would be useful, yet if a FREQ- result screen displays lamps without FREQ information it may be misconstrued as a non-FREQ- score. Obviously, this is also dependent on the skin creator's choice to include this information, however, I think LR2's decision to only show lamps when the score is 'savable' (outside of the generic Clear, Fail Blue/Red BG) makes sense and avoids this problem.

I prefer the former, providing a way to show FREQ setting while also displaying the clear type. Originally, I only implemented FREQ+ with the intention to avoid these complications as FREQ- is available in the practice tool and is redundant, I would consider that thinking outside the box.

seraxis:
> honestly, why do we even save FREQ scores at all? Does it make any sense to save FREQ+ scores with NO PLAY lamp? It could be made so you don't get your scores saved ever if FREQ is enabled, but all lamps would be displayed on result screen normally.

Many LR2 players set up a "FREQ player" and use exclusively that to track their freq scores, the current implementation supports this by saving scores locally (so you can track bp improvements etc) but doesnt submit to IR as there are fringe cases where freq+ is better than no freq, and also because Bokutachi's maintainer explicitly asked me not to.

As far as this PR is concerned I think I don't really understand the motivation to save fails specifically, and then have them overwritten at a later date with NOPLAY for a freq only local player account. Is this a use pattern that needs facilitating? Keeping this open for more discussion, I've already marked the referenced issue as wontfix but im interested in seeing how this develops.

MatVeiQaaa:
> > honestly, why do we even save FREQ scores at all? Does it make any sense to save FREQ+ scores with NO PLAY lamp? It could be made so you don't get your scores saved ever if FREQ is enabled, but all lamps would be displayed on result screen normally.
> 
> 
> 
> Many LR2 players set up a "FREQ player" and use exclusively that to track their freq scores, the current implementation supports this by saving scores locally (so you can track bp improvements etc) but doesnt submit to IR as there are fringe cases (omino charts, cyclone st10 among others) where freq+ is easier than no freq, and also because Bokutachi's maintainer explicitly asked me not to.
> 
> 
> 
> As far as this PR is concerned I think I don't really understand the motivation to save fails specifically, and then have them overwritten at a later date with NOPLAY for a freq only local player account. Is this a use pattern that needs facilitating? Keeping this open for more discussion, I've already marked the referenced issue as wontfix but im interested in seeing how this develops.

This does not make the game save fail lamps with freq. as I sometime described, saving and displaying a score is not connected and both can be done separately from one another, even with different data. This PR makes the game only display the fail lamp as requested in connected issue.
Then the question is: how should this behaviour end up looking like, if we keep developing this idea. Technically speaking anything imaginable could be done there without significant effort.

Catizard:
> > > Then the question is: how should this behaviour end up looking like, if we keep developing this idea. Technically speaking anything imaginable could be done there without significant effort.
> > 
> > 
> > Inspired from #8, we can store the freq data in a separate database file and dynamically set the data source when loading data from, `score.db`, for example. This should need much effort to work out, maybe we should rise another proposal/draft?
> 
> if you go that way, why bother with separate .db and inventing smart switching hacks, when you can just extend the existing one?

By adding fields like `freq` and `judge rank` to `score.db` directly is likely breaking the compatibility of returning endless dream to upstream. 

I'm kind of doubting the idea currently, maybe I need more thinking for a while :<

MatVeiQaaa:
> > > > Then the question is: how should this behaviour end up looking like, if we keep developing this idea. Technically speaking anything imaginable could be done there without significant effort.
> 
> > > 
> 
> > > 
> 
> > > Inspired from #8, we can store the freq data in a separate database file and dynamically set the data source when loading data from, `score.db`, for example. This should need much effort to work out, maybe we should rise another proposal/draft?
> 
> > 
> 
> > if you go that way, why bother with separate .db and inventing smart switching hacks, when you can just extend the existing one?
> 
> 
> 
> By adding fields like `freq` and `judge rank` to `score.db` directly is likely breaking the compatibility of returning endless dream to upstream. 
> 
> 
> 
> I'm kind of doubting the idea currently, maybe I need more thinking for a while :<

Extra fields don't break anything, it's the lack of fields that could be the problem, but we aren't doing that. Any fields are accessed by name, aren't they? Even then, you can make a separate table in the same DB and this would 100% not have even the theoretical a chance to break anything - separate DB is an unnecessary complication.

seraxis:
Separate table is the way to go. If you would like to update this PR with an implementation you have the green light. An ideal implementation would be one in which lamps and scores are only shown for the currently set freq value. But even a barebones results-screen only implementation would be fine.

seraxis:
This PR has been superseded by work in #103

Issue #27, opened 2024-03-27
Crash on random trainer enable


MatVeiQaaa:
The crash can be reproduced if you start the game, open random trainer menu, and enable "Black/White Random Select", then "Track Current Random", and then "Trainer Enabled". It occurs in method getLaneOrder() in file RandomTrainer.java at line 41, because laneOrder is null, so you get NullPointerException.

PR #28, opened 2024-03-29
Fix crash on first GetLaneOrder call with bw permute

MatVeiQaaa:
Fixes an issue #27 by initializing laneOrder variable. The crash occurs on first "Trainer Enable", if blackwhite permute was enabled, because the method GetLaneOrder is called before the first call to SetLaneOrder. If bw permute is enabled, GetLaneOrder will try to access the contents of laneOrder, which don't yet exist.


seraxis:
Nice catch, merged

PR #29, opened 2024-04-03
Allow quick restart with START or SELECT before playing a note

MatVeiQaaa:
Original behavior of beatoraja is to to run fade out animation and return to select screen, it's impossible to use quick restart feature with or without preserving the play options using START/SELECT in that state.

This changes that behavior, so you can restart the song in case you took longer than anticipated to drink water, check your stream chat, etc, without returning to select menu.

It achieves that by adding a new sub-state of playing state "STATE_ABORTED", which replaces STATE_PRACTICE_FINISHED, which contradictory to its name isn't necessarily ran in practice mode, for scenario where you quit before STATE_PRELOAD (loading resources stage) or STATE_READY (the stage where "get ready" skin animation plays) finish; and replaces STATE_FINISHED for scenario where you quit before the song ended without getting any judgements other than misses.

STATE_ABORTED reimplements a piece of code to restart the song with or without preserving the play options, as well as fade out animation and return to song select in case you just want to quit.

There is a non-zero possibility i missed some peculiar edge-case, where this change may cause some unexpected behavior, as this daisy chain of sub-states is rather difficult to follow. I am mostly confident i looked out for all possible scenarios and prevented possible issues.

### DIFF
--- a/core/src/bms/player/beatoraja/play/BMSPlayer.java
+++ b/core/src/bms/player/beatoraja/play/BMSPlayer.java
@@ -78,6 +78,7 @@ public class BMSPlayer extends MainState {
 	public static final int STATE_PLAY = 4;
 	public static final int STATE_FAILED = 5;
 	public static final int STATE_FINISHED = 6;
+	public static final int STATE_ABORTED = 7;
 
 	private long prevtime;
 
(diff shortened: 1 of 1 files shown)


seraxis:
Great feature, nothing stands out to me right now in your changes, merged

PR #34, opened 2024-05-18
Add configurations for easier linux debugging in IDEA

pfych:
I've modified the gradle configuration to append `-Djdk.gtk.version=2` when running from IDE since there are issues otherwise when working on Linux under Wayland.

I've also added run configs under `.idea` for debugging the created JAR


seraxis:
I don't think this should be the default configuration, this requires GTK 2 to be installed which may cause more confusion than it solves. If we could change the configuration dynamically based on environment that would be fantastic, or have some kind of set of developer packages

pfych:
I havent done a deep dive on the _actual_ issue yet but the UI in raja crashes on Wayland without that flag. With Wayland now being the default in a few mainline DEs this issue has started popping up. The AUR package for `lr2oraja-endlessdream` has gtk2 as a dependency for this reason 👍 

If you want I can clone `runShadow` & create `runShadowGTK2` as well as an `idea` run config for both GTK2 & without? 

Not too sure what workflow works best for you - I'm not a really java developer 😅 (Although I'm trying)

PR #43, opened 2024-10-30
Extend GetAssignedLane to BMSPlayer

marysiamzawka:
allows skins to fetch lane assignments during Play stage


seraxis:
Merged :+1:

PR #48, opened 2024-12-14
BugFix: Avoid constructing overlength sql

Catizard:
Problem: If you happen to have a very big BMS directory, this exception might occur:
```
严重: スコア取得時の例外:[SQLITE_TOOBIG] String or BLOB exceeds size
limit (statement too long) Query: SELECT * FROM score WHERE sha256 IN
(...)`
```

And the whole sqlite query statement would be printed in log file.

This commit fixed two overlength sql generate code place by forcing them to query data chunk by chunk. And one chunk size is set to be 1000.

### DIFF
--- a/core/src/bms/player/beatoraja/ScoreDatabaseAccessor.java
+++ b/core/src/bms/player/beatoraja/ScoreDatabaseAccessor.java
@@ -26,6 +26,7 @@ public class ScoreDatabaseAccessor extends SQLiteDatabaseAccessor {
 	private final ResultSetHandler<List<PlayerInformation>> infoHandler = new BeanListHandler<PlayerInformation>(PlayerInformation.class);
 	private final ResultSetHandler<List<ScoreData>> scoreHandler = new BeanListHandler<ScoreData>(ScoreData.class);
 	private final ResultSetHandler<List<PlayerData>> playerHandler = new BeanListHandler<PlayerData>(PlayerData.class);
+	private static final int LOAD_CHUNK_SIZE = 1000;
 
 	public ScoreDatabaseAccessor(String path) throws ClassNotFoundException {
 		super(new Table("info", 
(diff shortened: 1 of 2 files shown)


core/src/bms/player/beatoraja/ScoreDatabaseAccessor.java
```diff
@@ -157,17 +158,27 @@ public void getScoreDatas(ScoreDataCollector collector, SongData[] songs, int mo
...
+						str.append('\'').append(song.getSha256()).append('\'');
 					}
-					str.append('\'').append(song.getSha256()).append('\'');					
 				}
-			}
 
-			List<ScoreData> scores = Validatable.removeInvalidElements(qr
-					.query("SELECT * FROM score WHERE sha256 IN (" + str.toString() + ") AND mode = " + mode, scoreHandler));
+				List<ScoreData> subScores = Validatable.removeInvalidElements(qr
+						.query("SELECT * FROM score WHERE sha256 IN (" + str.toString() + ") AND mode = " + mode, scoreHandler));
```
corndogit, line 178: Just an observation for future investigation but I wonder if the logic being carried out in this whole method could be simplified by doing a join on the `sha256`, avoiding having to build a long `IN` clause?  Not claiming to be an expert in this though

corndogit (commented):
2 suggestions and a general comment

core/src/bms/player/beatoraja/ScoreDatabaseAccessor.java
Catizard: Yes, I was also curious about this issue a long time ago. An immature guess is that it is caused by the database being distributed and requiring `attach` before use. I will take another look at this after the `in-game download` and `reverse lookup` pr's work is done

corndogit (approved):

Catizard:
Thanks for the review! It's really hard for me to make decisions, approval makes me more confident <3

core/src/bms/player/beatoraja/ScoreDatabaseAccessor.java
Catizard: I did another investigation and found the invoke point is [here](https://github.com/seraxis/lr2oraja-endlessdream/blob/97303bd1ce17b32167763494d032365f5cfe2d1b/core/src/bms/player/beatoraja/select/BarManager.java#L796) and the parameter (which is too lengthy that causes the exception) `song` is generated from `bars`. 

Some of the bars are technically able to rewrite the long `IN` clause with `join`. For example, if a bar represents one bms source directory on disk, we can attach and generate the sql that joins the `song` table with a where clause that filters the parent directory. However, it'll be impossible for a bar that represents a difficult table or a difficult table's level folder since data are stored on disk.

Therefore, I personally think the idea is great but it cannot simplify the code structure currently. We might need much more rewrite to make this idea come true.

PR #58, opened 2025-03-07
Fix post-merge bug where scores woudn't send

MatVeiQaaa:
merge with 0.8.8 caused an issue, creating a second score send queue, which was partially used, while the previous queue was also used partially... resulting in no work being done in total (lol)

Also reduces the wait time before one can exit the result scene from 3 seconds to 1 second.

### DIFF
--- a/core/src/bms/player/beatoraja/result/MusicResult.java
+++ b/core/src/bms/player/beatoraja/result/MusicResult.java
@@ -33,8 +33,6 @@ import bms.player.beatoraja.song.SongData;
 public class MusicResult extends AbstractResult {
 
 	private ResultKeyProperty property;
-	
-	private List<IRSendStatus> irSendStatus = new ArrayList<IRSendStatus>();
 
 	public MusicResult(MainController main) {
 		super(main);
@@ -107,8 +105,12 @@ public class MusicResult extends AbstractResult {
 				int irsend = 0;
 				boolean succeed = true;
 				List<IRSendStatus> removeIrSendStatus = new ArrayList<IRSendStatus>();
+				List<IRSendStatus> scores = new ArrayList<IRSendStatus>();
+				if (!main.irSendStatus.isEmpty()) {
+					scores = main.irSendStatus.subList(main.irSendStatus.size() - ir.length, main.irSendStatus.size());
+				}
 
-				for (IRSendStatus irc : irSendStatus) {
+				for (IRSendStatus irc : scores) {
 					try {
 						if (irsend == 0) {
 							timer.switchTimer(TIMER_IR_CONNECT_BEGIN, true);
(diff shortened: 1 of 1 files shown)


MatVeiQaaa:
> Hi @MatVeiQaaa
> 
> 
> 
> This issue still exists on my machine, after reading your pr and code I found the following problem:
> 
> 
> 
> - The ir send thread from `MainController` iterates through ir connections correctly but no one satisfy that `score.retry !=0` since no score has already been sent
> 
> - `irSendStatus` inside `MusicResult` is empty because [here](https://github.com/seraxis/lr2oraja-endlessdream/blob/319128e5c41dc042d0f45792cdf06c520bb5ca8f/core/src/bms/player/beatoraja/result/MusicResult.java#L101) only adds ir status to `MainController`
> 
> 
> 
> I opened a draft pr #64 which simply adds the ir status to `MusicResult` one and ir works on my machine now :)

I think what happens is it might only send the score to the main IR, when you got more than one IR connected. You can see a bit lower in the code it creates an IR send thread in MusicResult, creating and sending a score list from a subset of queue at MainController.

Sadly I don't have access to a computer right now and can't properly look into it, but that's likely what happened.

Catizard:
> > Hi @MatVeiQaaa
> > This issue still exists on my machine, after reading your pr and code I found the following problem:
> > 
> > * The ir send thread from `MainController` iterates through ir connections correctly but no one satisfy that `score.retry !=0` since no score has already been sent
> > * `irSendStatus` inside `MusicResult` is empty because [here](https://github.com/seraxis/lr2oraja-endlessdream/blob/319128e5c41dc042d0f45792cdf06c520bb5ca8f/core/src/bms/player/beatoraja/result/MusicResult.java#L101) only adds ir status to `MainController`
> > 
> > I opened a draft pr #64 which simply adds the ir status to `MusicResult` one and ir works on my machine now :)
> 
> I think what happens is it might only send the score to the main IR, when you got more than one IR connected. You can see a bit lower in the code it creates an IR send thread in MusicResult, creating and sending a score list from a subset of queue at MainController.
> 
> Sadly I don't have access to a computer right now and can't properly look into it, but that's likely what happened.
> 
> Also, check what code you were working with in your draft, as irSendStatus object shouldn't exist at all in MusicResult class, as I removed it in this PR. You might have been lacking the code of this PR. Or maybe the merge into 0.3.0 branch was wrong.

Yes thank you MatVeiQaaa, my local branch code is missing exactly one commit from `0.3.0` and it's this pr, currently I still need a patch to make it runnable on macos, which makes my local branch is a little messy and causes this unfortunate misunderstanding. I'm really sorry :(
I have tested again and ir score sending is working as expected, sorry to bother T_T

PR #61, opened 2025-04-09
Fix broken MS-ADPCM decoding implementation

### LINKED ISSUE #37: [Request] support for ADPCM encoded WAV files
PMS files converted from arcade use these.
lr2 is able to play these files properly, but in beatoraja they end up silent.

Merrg1n:
Close #37

I've tested on these BMS that use MS-ADPCM:
* Romancing Sa・Ga3 ～四魔貴族バトル2～
* エストポリス伝記II ～バトル# 3～
* フランの中できゅっ☆としてドカーン†

All these WAVs are stereo. I can't find any WAV files with more than two channels to test with, but the handling of this part of the code should be consistent with ffmpeg's implementation.

### DIFF
--- a/core/src/bms/player/beatoraja/audio/MSADPCMDecoder.java
+++ b/core/src/bms/player/beatoraja/audio/MSADPCMDecoder.java
@@ -1,10 +1,9 @@
 package bms.player.beatoraja.audio;
 
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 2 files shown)


Merrg1n:
Since some older BMS files also use MS-ADPCM, perhaps we should also submit this change to upstream?

seraxis:
Wow! thank you for fixing this, i left it in a broken state as there were some bugs i couldn't figure out at the time.

PR #62, opened 2025-04-12
feat: open full-screen mode at specified monitor

Catizard:
This commit fixes the issue that user open full-screen mode while specify the monitor to use.

> I will refactor the code as soon as I find some time

## Details

`libgdx` doesn't provide an api allows us to open on specified monitor, but provides an api allows us speicfy the window position before opening. This commit's appoarch is first open the window in `boardless` mode and move the window to the speicifed monitor. Then once the window is created, entering full-screen mode.

> Why not `windowed` mode but `boardless` mode?
> `windowed` mode ships with a `title bar`, which would make the 
window's content moves up after entering full-screen, it looks like some of the game's content is being cut out.

## Drop-in Compatebility

This commit handles the cases that `monitor` option is empty, cannot be found to keep the drop-in compatebiliity. The only possible break is the upstream beatoraja also provides `monitor` option one day, which may have different definition that conflicts the one from this pr.

### DIFF
--- a/core/src/bms/player/beatoraja/Config.java
+++ b/core/src/bms/player/beatoraja/Config.java
@@ -164,6 +164,7 @@ public class Config implements Validatable {
 
 	private boolean useDiscordRPC = false;
 	private boolean setClipboardScreenshot = false;
+	private String monitorName = "";
 
 	private static final String[] DEFAULT_TABLEURL = { "http://bmsnormal2.syuriken.jp/table.html",
 			"http://bmsnormal2.syuriken.jp/table_insane.html",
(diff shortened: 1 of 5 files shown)


Catizard:
Side note: this pr seems impossible for upstream beatoraja if upstream is still using `lwjgl2`

Catizard:
This pr has some relevance to #25 and does part of the job it requires, should we solve two issues together?
It wouldn't be very hard to enumerate all possible resolution but resolution config value ships from upstream is using string constants like `HD`, which is a little bit annoying to maintain drop-in compatibility.

seraxis:
Once the merge conflicts on this are resolved we can merge

PR #63, opened 2025-04-19
Fix build error in MSADPCMDecoder.java

llm96:
Very minor fix for a build error in #61 - could do with a quick test in-game as I don't have any affected charts on hand.

Build test run: https://github.com/llm96/lr2oraja-endlessdream/actions/runs/14550170295


Merrg1n:
Oh, that's my mistake. My local build environment is JDK 17, and I didn't realize that build action is using JDK 8.

PR #64, opened 2025-05-09
fix: ir score is still wouldn't send

Catizard:
`irSendStatus` from `MainController` one is a retry thread, the initial one is located at `MusicResult`. However it does nothing because its `irSendStatus` is always empty. This commit fills the empty.


Catizard:
It's a mistake from my local branch, #58 has fix everything correctly

PR #65, opened 2025-05-29
Hidden enable/disable on SELECT, change setCoverValue priority

MatVeiQaaa:
This allow enabling/disabling hidden cover from play scene by double press of SELECT, as well as changing every value of SUD, HID and LIFT without having to exit the scene, by changing the priority of HID vs LIFT when holding START. This way you don't need to disable lift, which is only doable from outside the scene, to change hidden cover value, but lift can be adjusted by disabling hidden cover with double press of SELECT.


MatVeiQaaa:
@Catizard for discovery, there must be a "cheat-sheet" similar to what LR2 has, describing all the "hidden" features, in-game. Iirc in LR2 that would be a menu brought up with F3 button (or F1?). Not even upstream advertises its hidden keybinds in any way to my knowledge.

PR #67, opened 2025-06-07
Upgrade Actions version and JDK Version

Lighthigh57:
Since some builds were failing due to the use of an older JDK, we have changed to use JDK17, which is also a minimum requirement for beatoraja. Also, Actions used in workflow have been brought up to date.

This PR message uses DeepL for translation.


Catizard:
Going to close this since it's duplicated and  #63 is preferred

PR #69, opened 2025-07-29
Add `SongManager` menu for rendering current reverse table lookup

### LINKED ISSUE #26: [Request] Reverse Folder Lookup
QoL Feature to be able to see what tables a chart is in by pressing a button.

Catizard:
close #26 

This commit adds a new menu called `Song Manager`, which is designed for rendering current song's status and providing easy access to some operations (e.g. reverse table lookup & favorite folder management).

As for now, the menu shows the `best record set time` as a "current song's status" example and `reverse table lookup` button as an "extra operations" example.

For future reviewers: this pr picks a temporary solution for adding the difficult tables' symbols, it can be deleted after ImGUI-java upgraded it's dependency version to 1.92 or above. See below comments for details.

### DIFF
--- a/core/src/bms/player/beatoraja/PlayerResource.java
+++ b/core/src/bms/player/beatoraja/PlayerResource.java
@@ -2,7 +2,7 @@ package bms.player.beatoraja;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.HashMap;
+import java.util.*;
 import java.io.FileInputStream;
 import java.io.ObjectInputStream;
 import java.io.IOException;
@@ -127,6 +127,8 @@ public final class PlayerResource {
 	private String tablename = "";
 	private String tablelevel = "";
 	private String tablefull;
+	// Full list of difficult tables that contains current song
+	private List<String> reverseLookup = new ArrayList<>();
 
 	public PlayerResource(AudioDriver audio, Config config, PlayerConfig pconfig) {
 		this.config = config;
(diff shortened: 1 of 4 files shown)


core/src/bms/player/beatoraja/modmenu/ImGuiRenderer.java
```diff
@@ -59,6 +59,7 @@ public static void init() {
         rangesBuilder.addRanges(io.getFonts().getGlyphRangesDefault());
         rangesBuilder.addRanges(io.getFonts().getGlyphRangesCyrillic());
         rangesBuilder.addRanges(io.getFonts().getGlyphRangesJapanese());
+        rangesBuilder.addChar('★');
```
Catizard, line 62: ImGUI requires adding the glyph ranges manually. Currently, I didn't find a good solution for adding ranges that contains the difficult table symbols

core/src/bms/player/beatoraja/PlayerResource.java
```diff
@@ -548,6 +550,36 @@ private void setTableinfo(){
 		setTablelevel("");
 	}
 
+	public List<String> getReverseLookupData(String md5, String sha256) {
```
Catizard, line 553: Beatoraja stores difficult table data as binary data on disk, which cause reverse lookup is a very heavy job to do. Due to the ImGUI's implementation strategy, it seems impossible to trigger the reverse lookup "exactly once" when user clicking the button. 
Another options is adding a "threshold": trigger the reverse lookup once after a period of time without switching song.

core/src/bms/player/beatoraja/modmenu/ImGuiRenderer.java
Catizard: Dear ImGUI 1.92 improves the glyph settings: manual glyph ranges setup is no longer needed. I think we can add all common difficult table symbols as a temporary solution for this and remove them when ImGUI-java has upgraded to `1.92` or above

PR #71, opened 2025-08-08
Persist some freq related data to avoid changing freq during gameplay

### LINKED ISSUE #45: (Bug) Disabling rate mod mid-play submits score
Disabling or changing the rate to 100% or above saves and submits the score to IR despite the song being played at the rate it was originally set at. 
It is possible to play the song at 50% speed, disable the rate mod, and have it count as a legitimate clear.

Related bug where the title changes depending on what the rate mod is currently set to and not where it started at. For example, you can start the song at 50% rate, change the rate to 150%, and the song title will show [1.50x] instead of [0.50x]

Catizard:
close #45 

This pr should fix two problems that pointed out:

- Disabling the freq trainer will confuse endless dream and think it's a normal score and submit it to IRs
- Title will be changed when changing freq during the game play

by persisting some "flags/data".

This pr intentionally leaves the local score save related code unchanged. Also the clear type won't be NO_PLAY if the user used the trick above.

### DIFF
--- a/core/src/bms/player/beatoraja/PlayerResource.java
+++ b/core/src/bms/player/beatoraja/PlayerResource.java
@@ -1,15 +1,5 @@
 package bms.player.beatoraja;
 
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.util.*;
-import java.io.FileInputStream;
-import java.io.ObjectInputStream;
-import java.io.IOException;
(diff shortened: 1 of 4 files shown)


Catizard:
This is a possible cheat way, since the result is being considered as legitimate clear. Maybe after 0.3.0 release we need to connect IRs to stop supporting previous versions

Issue #23, opened 2024-03-19
Option to skip decide screen


corndogit:
Small QoL feature, this would add the option to skip the transition between select screen and play screen.

Currently you can press button 1 on the decide screen to skip the decide screen animation (after about a second) so this option is mostly just to automate that without any extra user input.

marysiamzawka:
> Currently you can press button 1 on the decide screen to skip the decide screen animation (after about a second)

The time it takes for a skin to fade out or start accepting inputs is definiable in the skin's files. You can achieve the result you're looking for by creating a decide skin that lasts 1 milisecond.

seraxis:
I'm half remembering this but I think completely skipping the decide state could lead to a possible race condition in the music thread so I'm afraid the skin solution will have to do for now, keeping this open though

MatVeiQaaa:
> I'm half remembering this but I think completely skipping the decide state could lead to a possible race condition in the music thread so I'm afraid the skin solution will have to do for now, keeping this open though

Realistically speaking nothing would happen, you either switch to next scene before starting the sound, or stop the sound before switching the scene, there is a method for that. It is likely you can even skip switching to decide scene all together and go straight into play scene, i don't imagine decide scene performing any crucial operations. However i find this weird and unnecessary, i don't want to be the one to make that.

corndogit:
Added in bf96553e7e01ebbe857af92ab5451e8677c8170b

PR #88, opened 2025-09-23
#23: Option to skip decide screen

corndogit:
Addresses #23. This PR adds a config option under the "Music Select" tab in the launcher to allow skipping the decide screen when selecting a song.

- This option is disabled by default in the launcher
- When selecting a song, the game switches directly to the play scene instead of switching to the decide scene, skipping the animation and decide BGM sound.

### Skip decide screen enabled

https://github.com/user-attachments/assets/140b7c78-855c-485c-bb96-1d7bfbe830b7

### Skip decide screen disabled (default)

https://github.com/user-attachments/assets/72e4fd5b-c6a6-47aa-9447-7c1e09e05cf8

### DIFF
--- a/core/src/bms/player/beatoraja/Config.java
+++ b/core/src/bms/player/beatoraja/Config.java
@@ -72,6 +72,10 @@ public class Config implements Validatable {
 	 */
 	private int maxSearchBarCount = 10;
 	/**
+	 * When selecting a song, if set to true, game changes to play scene instead of decide scene
+	 */
+	private boolean skipDecideScreen = false;
+	/**
 	 * 所持していない楽曲バーを表示するかどうか
 	 */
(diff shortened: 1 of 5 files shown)


Catizard (approved):
There's only one thing looks suspicious to me, [`MusicScene` modifies a field in `resource`](https://github.com/seraxis/lr2oraja-endlessdream/blob/8a8f5c018808307ae1a1b408bba6f95fc9351829/core/src/bms/player/beatoraja/decide/MusicDecide.java#L29)
However, this field seems to be a dead feature because the field can never have any different value than [its initial value](https://github.com/seraxis/lr2oraja-endlessdream/blob/8a8f5c018808307ae1a1b408bba6f95fc9351829/core/src/bms/player/beatoraja/PlayerResource.java#L139) so I think it's fine.

PR #94, opened 2025-09-24
TT emblem rotation

Kolyasisan:
* Reimplemented TT emblem rotation on Play screen
* Now behaves more like IIDX (faster, smoother motions with momentum, less rigid)

Example:
https://github.com/user-attachments/assets/0a6392c1-6bd7-41f1-9c16-a7a61d9eb8a0


seraxis:
Merged

PR #98, opened 2025-09-28
Launch the game in a new thread, to avoid locking up JavaFX

Arctice:
On Linux (and possibly MacOS), the `MainLoader.play` call in PlayConfigurationView:721 causes the JavaFX thread to essentially become the main game thread, but doing so locks up the launcher and causes deadlocks in specific calls, in particular:
- pressing F3 while hovering a song (`open_with_explorer` event at EventFactory.java:311) deadlocks within `Desktop.getDesktop().open()`, at `gnome_url_show`
- bokutachi IR's startup warning message, with the widely distributed `GTKLookAndFeel` envvar and no `SHUT_UP_TACHI`, causes another freeze

To prevent this, we detach the `play` call into a new thread, letting the JavaFX thread progress. This appears to fix the above issues but raises some questions that might need testing on other platforms:
- How does this work on Windows without freezing up the launcher in the first place (I'm guessing jwgl creates its own thread)? Can putting play in a separate thread potentially cause race conditions?
- Does this fix work on MacOS?


Arctice:
Crashes on mac, need to switch to a different approach or platform specific code

PR #100, opened 2025-09-29
I18n support for notifications

Catizard:
This is a proposal for how we implement the i18n support for the in-game notifications:

- Reuse the `UIResources.properties` & `ResourceBundle` mechanism. This commit adds 3 messages for example, one is a simple welcome message, two are IR send success/failed messages.
- Introducing a simple structure to referencing the message key located at property files. See [here](https://github.com/Catizard/lr2oraja-endlessdream/commit/17c9742d9a63c82a686b43870c8df3ec2da581f7#diff-9b9a3a7e420bdcdac385a52792de9eab6d43522aa43511dd7a5cedc7c49d34c3R16-R23). Although here's only one level, but it can be nested for more complex modules.


Catizard:
I uploaded a change as a new proposal for error messages, which includes a way to automatically setup and print error code. Demanding further discussion.

PR #103, opened 2025-10-02
Store freq & override judge score in separate table

Catizard:
This is a proposal of how we store the scores have endless dream's features like freq/override judge. This pr creates a similar table `eddatalog` in `scoredatalog.db`, but:

- Removes the primary key, which causes `scoredatalog.db` file doesn't save each play record
- Adds two new fields `rate` and `overridejudge`

Needs further discussion:

- Should we transfer the old data from table `scoredatalog`?
- How should we treat `score.db` and `scorelog.db`?

### DIFF
--- a/core/src/bms/player/beatoraja/PlayerResource.java
+++ b/core/src/bms/player/beatoraja/PlayerResource.java
@@ -128,7 +128,9 @@ public final class PlayerResource {
 	private String tablelevel = "";
 	private String tablefull;
 	private boolean freqOn;
+	private int freqValue;
 	private String freqString;
+	private int overrideJudgeRank;
 	// Full list of difficult tables that contains current song
 	private List<String> reverseLookup = new ArrayList<>();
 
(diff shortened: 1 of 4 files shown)


Catizard:
Fixed the lamp folder display issue and fill the fallback strategy. This pr is now can be considered as completed. But not ready to be merged, demanding further discussion and fill the todos

Catizard:
This pr allows users to view their scores on courses too:

image

During implementing this feature, I found a strange thing is raja actually doesn't record course result into `scoredatalog`. So I filled this behavior in this pr, which may cause conflicts in the future.

Catizard:
image

This pr has an uncommitted extension, based on #179, which allows user review their score in result scene/course result scene. I couldn't decide whether to include this feature in this pr or not. Need some advice :)

Catizard:
Update: This pr is now rebased onto #206 since it removes the `SongManagerMenu`. Converting to draft again.

image

PR #105, opened 2025-10-06
Implement per-chart volume adjustment

llm96:
Upon selecting a chart, mixes all sounds into a PCM buffer, run it through libebur128 to analyze how loud it is, then adjusts the Key and BG volumes accordingly. Analysis is performed once, then written to disk cache for future use.

Results are briefly displayed before the chart starts. This was mostly for debugging purposes and should probably be removed later. The feature is disabled by default. Configurable from the Audio tab in the launcher.

image

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -95,6 +95,7 @@ dependencies {
     implementation(libs.guacamole)
 
     implementation(libs.ebur128java)
+    implementation(libs.jna)
 
     // non-gradle managed file dependencies. jportaudio not on maven. "custom" scares me.
     implementation(":jportaudio")
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -20,6 +20,7 @@ xz = "1.10"
 shapedrawer = "2.5.0"
 guacamole = "0.3.5"
 ebur128java = "1.2.6-2"
+jna = "5.13.0"
 
 
 [libraries]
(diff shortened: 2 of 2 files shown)


Catizard:
Breaks on arm macos. It seems that the jna is confused when there're multiple library files that have same name. Jna indicates that the library it tries to load is an i386 architecture file, but wants an arm instead. Currently I haven't figured out a way to handle this problem, here's some thoughts:

- If we can specify which file to load into, we might be able to detect current platform and load it manually?
- Imgui/lwjgl is providing native library files as a standalone dependency and we filter the dependencies manually:

https://github.com/seraxis/lr2oraja-endlessdream/blob/1c6923d36e9dd915d684385f29b338b244acc3ef/buildSrc/src/main/kotlin/org/endlessdream/extra/PlatformFilter.kt#L37-L54
However, this seems overkill.

Catizard:
>Can you instead try building with this patch? I think the current ebur128java version on Maven should work, but if not, the one in my previous comment with the Universal binary should definitely do it.

After applying the patch everything works.

>Nevermind, I had the wrong idea from the log message. The error seems to be related to the JNA dispatch library instead.
>
>It looks like the Discord RPC package uses a really old version that predates macOS ARM support. I noticed the binary for it >included in the latest 0.3.0 version is actually an x86 one.
>
>Does that feature work as expected for you? I suspect it would crash in a similar way if enabled.

Yeah it would crash with an identical exception message, we have noticed this issue days ago but misunderstood it as a library file issue rather than jna issue :<

PR #113, opened 2025-10-11
update java-discord-rpc library's jna dependency

Catizard:
#105 requires `jna` as a dependency, therefore, there're two libraries in endless dream have jna dependency, the other one is `java-discord-rpc`. Unfortunately, the `jna` dependency included by `java-discord-rpc` is old and it doesn't support arm macos, which causes immediately crash if accessed jna library.

Somehow, if we reorder the dependencies order, i.e. placing the statement declaring the jna dependency before discord, everything works fine. But cannot reversely. This is the reason why at code review stage I confirmed #105 has no conflict issue on arm macos. However, this is related to the code order and currently it's broken on macos.

However, solving this issue by simply reordering the dependencies order sounds fragile. Therefore, this pr updates binary jar included in endless dream repo, which is built from [my personal fork](https://github.com/Catizard/java-discord-rpc) and has the same version(5.13.0) jna dependency. This is a stopgap measure until we can fully replace this dependency


seraxis (approved):
Hopefully we can work out a better fix for this, merged

PR #122, opened 2025-10-13
Simple notification when assist option is on or score won't be recorded

### LINKED ISSUE #116: [Feature] Notify users when their current options will not save scores
When a user has constant, remove LN, or other combination of options that disables score or IR submission there should be either an info popup or something else to let them know that their play will not save.

Catizard:
close #116 

~~There's an [extra case](https://github.com/seraxis/lr2oraja-endlessdream/blob/3044baf958826f69b8eded5e47107b32ad88b801/core/src/bms/player/beatoraja/result/MusicResult.java#L454C1-L461C4) cannot be captured.~~ Changed in this pr.

### DIFF
--- a/core/src/bms/player/beatoraja/play/BMSPlayer.java
+++ b/core/src/bms/player/beatoraja/play/BMSPlayer.java
@@ -292,6 +292,12 @@ public class BMSPlayer extends MainState {
 				model.setJudgerank(overridingJudgeWindowRate);
 			}
 
+			// Constant considered as assist in Endless Dream
+			// This is a community discussion result, see https://github.com/seraxis/lr2oraja-endlessdream/issues/42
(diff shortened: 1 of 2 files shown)


core/src/bms/player/beatoraja/play/BMSPlayer.java
```diff
@@ -420,6 +420,12 @@ public BMSPlayer(MainController main, PlayerResource resource) {
 			gaugelog[i] = new FloatArray(playtime / 500 + 2);
 		}
 
+		if (assist != 0) {
+			ImGuiNotify.warning("Next play would be considered as assist clear");
+		}
+		if (!score) {
+			ImGuiNotify.warning("Next play score wouldn't be recorded");
```
seraxis: Perhaps reword to 
"Assist options enabled. Next play will be saved as an assist clear"
"Score nullifying options enabled. Next play will not be saved"
respectively

PR #123, opened 2025-10-13
Add Lombok to the project

corndogit:
## Description

This PR adds [Lombok](https://projectlombok.org) as a dependency to the project, allowing us to replace a lot of boilerplate code with autogenerated methods, aligning the codebase with more modernised Java practices and work towards improving the readability and maintainability of the codebase.

## A brief background on Lombok 

Lombok is a library for Java that adds compile-time annotations for auto-generating code, making it easier and quicker to build classes that are easier to maintain and containing less clutter.

To give an example with Lombok

```java
@Getter
@Setter
@AllArgsConstructor
public class Foo {
  private String myField;

  @Setter(AccessLevel.NONE)  // Disables generation of a setter for this field
  private boolean myBool;
}
```

After compilation, the constructor, getter and setter methods would be generated for the field `myField` like this:

```java
public class Foo {
  private String myField;
  private boolean myBool;

  public Foo(String myField, boolean myBool) {
    this.myfield = myField;  
    this.myBool = myBool;
  }

  public String getMyField() {
    return this.myField;
  }

  public void setMyField(String myField) {
    this.myfield = myField;  
  }

  public boolean isMyBool() {
    return this.myBool;
  }
}
```

*For further reading, I recommend looking at [Baeldung](https://www.baeldung.com/intro-to-project-lombok) and the Lombok documentation linked in this PR*

Intellij's Lombok plugin supports identifying methods that would be generated by Lombok annotations, making them available for code autocompletion, as well as a "delombok" tool that can be run to replace all annotations with their generated code equivalents (useful for debugging)

As a bit of a proof-of-concept, the `PlayerConfig` class has been refactored to remove getters and setters, relying completely on those generated by `@Getter` and `@Setter` where additional logic is not required. This alone results in ~47% line count reduction in pre-compilation source code from 1152 lines of code to 591 lines, with further improvements still possible.

## Still todo (more points to come as I'm working on it)

- More testing to ensure nothing is broken
  - Played on the `shadowJar` for a bit and seemed fine but want to triple-check since breaking player config would be disastrous
- Under this PR I plan to add documentation to convey that Lombok can be used within new code to replace otherwise boilerplate code

## Future work

- Explore [other annotations](https://projectlombok.org/features/) available in Lombok that can be used e.g. `@Data`, and define guidance on what not to use
  - Despite the amount of try-catch for checked exceptions, I think we shouldn't use `@SneakyThrows` in most cases as it encourages bad error handling. We should instead focus on improving the handling of these cases, providing better logging and ensuring we keep the application in a healthy state
- Create follow-up PRs to apply Lombok to other config files


corndogit:
After discussing with @seraxis we decided this would cause too many major changes throughout classes. With the additional overhead of having to rename many variables and methods throughout different parts of the code to comply with Lombok's naming conventions, there is a risk of breaking stuff unknowingly with a large scope for testing to ensure this doesn't happen. Closing the PR

PR #125, opened 2025-10-14
Show user name in lr2ir leaderboard & highlight

Catizard:
This pr aims to fix two issues currently the LR2IR leaderboard has:

- The local score is showing as a blank. Replaced by user's name in this pr
- Giving the local score bar a different bar type to highlight it

This pr can use `FunctionBar` to remove the `IRPlayerBar`, but needs some changes to make it possible.


Arctice (changes requested):
Currently does not compile due to leftover references to previous implementation, and FunctionBars don't support rendering customized scores the way IRPlayerBars did

seraxis (approved):
great stuff, merged

PR #128, opened 2025-10-17
Fix: crash in database update for some folders of the AI table 

Arctice:
Fixes a catastrophic bug. The offending code filters some elements from a list of songs, then converts it into an array pre-allocated with a size of this list from before the removals, thus ending up with nulls at the end of the array.


seraxis (approved):
Thanks for this one, merged

PR #135, opened 2025-10-23
feat: OBS WebSocket Control

NotAkitake:
This PR aims to bring OBS WebSocket control based on the in-game state, featuring:

- OBS scene switching
- OBS actions (Start/Stop Recording)
- Handles quick retries (will restart recording if ongoing)
- Reconnects to the WebSocket if connection is lost
- Multiple recording file handling methods:
    - Let OBS keep all files (do nothing)
    - Keep the recording if a screenshot was taken (on screenshot) , otherwise delete on next recording start
    - Keep the recording file if the play was saved to a replay slot (1-4 pressed on result), otherwise delete on next recording start
- In-game notifications on recording start/stop
- Customizable delay before a recording is stopped. Note that if the game state changes again and a StopRecord action is due, it'll run it right away

Reminder this is my first time working with java and this codebase, any feedback or advice is appreciated.

### DIFF
--- a/core/src/bms/player/beatoraja/MainState.java
+++ b/core/src/bms/player/beatoraja/MainState.java
@@ -190,16 +190,5 @@ public abstract class MainState {
 	
 	public enum MainStateType {
 		MUSICSELECT,DECIDE,PLAY,RESULT,COURSERESULT,CONFIG,SKINCONFIG;
-
-		public static MainStateType fromString(String name) {
-			if (name == null)
-				return null;
-			for (MainStateType state : MainStateType.values()) {
-				if (state.name().equalsIgnoreCase(name)) {
-					return state;
-				}
-			}
-			return null;
-		}
 	}
 }
--- a/core/src/bms/player/beatoraja/obs/ObsWsClient.java
+++ b/core/src/bms/player/beatoraja/obs/ObsWsClient.java
@@ -74,15 +74,6 @@ public class ObsWsClient {
 			return value;
 		}
 
-		public static ObsRecordingMode fromString(String str) {
-			for (ObsRecordingMode mode : ObsRecordingMode.values()) {
-				if (mode.name().equals(str)) {
-					return mode;
-				}
-			}
-			throw new IllegalArgumentException("No matching enum for value: " + str);
-		}
-
 		public static ObsRecordingMode fromValue(int value) {
 			for (ObsRecordingMode mode : ObsRecordingMode.values()) {
 				if (mode.getValue() == value) {
@@ -444,7 +435,7 @@ public class ObsWsClient {
 			return;
 		}
 
-		final ObsRecordingMode reasonMode = ObsRecordingMode.fromString(reason);
+		final ObsRecordingMode reasonMode = ObsRecordingMode.valueOf(reason);
 		if (this.saveRequested || recordingMode == ObsRecordingMode.KEEP_ALL || reasonMode != recordingMode) {
 			return;
 		}


Catizard:
> As for the connect button, you only need to connect to change your scene/actions. If you already have them set you can just hit play. Another WS client is created when you start the game (in ObsListener, created in MainController). Also your state choices are only committed to config if you connected in the launcher, otherwise it skips it in commit (see saveSelections method)

This design is kinda counterintuitive. There's a connect button and everything is disabled before clicking the connect button. I'll check if I can tweak this after I look into the code later.

> I'm not sure what you mean by "Storing replay" there at the end, could you explain in more detail?

Persisting the video when user storing the replay (pressing num1/2/3/4 at result screen) like `On screenshot` option.

core/src/bms/player/beatoraja/obs/ObsWsClient.java
```diff
@@ -0,0 +1,603 @@
...
+
+				if (reconnectTask != null && !reconnectTask.isDone()) {
+					reconnectTask.cancel(false);
+					reconnectTask = null;
+				}
+
+				System.out.println("Connected to OBS WebSocket");
+
+				if (onOpenHandler != null) {
+					onOpenHandler.run();
```
Catizard: Unused hook?

core/src/bms/player/beatoraja/obs/ObsWsClient.java
```diff
@@ -0,0 +1,603 @@
...
+					onErrorHandler.accept(ex);
+				}
+			}
+		};
+	}
+
+	private void handleEvent(JsonNode json) {
+		try {
+			JsonNode d = json.get("d");
+			if (d != null && d.has("eventType") && d.has("eventData")) {
```
Catizard: Suggestion: quick return if this is an invalid message
```java
if (d == null || !d.has("eventType" || !d.has("eventData") {
    return ;
}
```

core/src/bms/player/beatoraja/obs/ObsWsClient.java
```diff
@@ -0,0 +1,603 @@
...
+				}
+
+				createWebSocketClient();
+
+				if (!wsClient.connectBlocking(5, TimeUnit.SECONDS)) {
+					throw new Exception("Connection timeout");
+				}
+
+			} catch (Exception e) {
+				currentReconnectDelay = Math.min(
```
Catizard: Maybe better to add a note here: reconnect delay would be reset once successfully connected

core/src/bms/player/beatoraja/result/MusicResult.java
```diff
@@ -326,6 +326,7 @@ public void saveReplayData(int index) {
 				main.getPlayDataAccessor().wrireReplayData(rd, resource.getBMSModel(),
 						resource.getPlayerConfig().getLnmode(), index);
 				saveReplay[index] = ReplayStatus.SAVED;
+				main.saveLastRecording();
```
Catizard: This needs to be wrapped with config check. Currently either saving replay or taking a screenshot both can keep the last record no matter whether the option is `On screenshot` or `Saving replay`

core/src/bms/player/beatoraja/Config.java
```diff
@@ -704,6 +752,56 @@ public void setWebhookUrl(String[] webhookUrl) {
...
+	public int getObsWsRecStopWait() { return obsWsRecStopWait; }
+
+	public void setObsWsRecStopWait(int obsWsRecStopWait) {
+		this.obsWsRecStopWait = Math.max(0, Math.min(obsWsRecStopWait, 10000));
+	}
+
+	public int getObsWsRecMode() { return obsWsRecMode; }
+
+	public void setObsWsRecMode(int obsWsRecMode) {
+		this.obsWsRecMode = Math.max(0, Math.min(obsWsRecMode, 2));
```
Catizard: I don't suggest to modify the value without leaving a sign

Catizard (changes requested):
Only the `MusicResult` & `CourseResult`'s review is requesting changes. Others are just suggestions.

core/src/bms/player/beatoraja/obs/ObsWsClient.java
NotAkitake: Yup, I usually do this but didn't catch this one, makes sense to indent less.

core/src/bms/player/beatoraja/obs/ObsWsClient.java
NotAkitake: Not sure what you mean by this. You mean a comment?

core/src/bms/player/beatoraja/result/MusicResult.java
NotAkitake: Yeah good catch, I actually had done this but then removed it, forgot why.
Made it pass a `reason` string that gets interpreted as ObsRecordingMode and checks if it matches the current recording mode.

Arctice:
Apologies for not using the github review builder, I find it difficult to work with. This is a review of the PR's state at c14b7940.

So far I've only been able to review some of the changes introduced in this PR - the ones in existing files and `ObsListener.java`; I've not looked at the JavaFX parts nor the much more complicated `ObsWsClient.java`.


`Config.java:203-204`
```java
private Map<String, String> obsScenes = new HashMap<>();
private Map<String, String> obsActions = new HashMap<>();
```
Because the new `Config` members are abstract `Map`s, this appears to insert a `"class": "java.util.HashMap"` entry in our `config_sys.json` file. Please keep in mind that those config files are intended to be user-readable and leaving in entries such as these is liable to create confusion at worst and looks bad at best. I understand the instinct to use `Map` member variables over the concrete `HashMap`, but in this case I'd suggest using the latter.


`MusicResult.java:329`
I don't fully understand all of the checks that occur at lines 322-324, but at a glance, making `saveLastRecording` depend on them seems odd - for example, if I understand the admittedly-horrifying `BMSPlayer` constructor correctly, the following block of code won't be entered if the score from the current play won't be saved due to custom judge or another similar option. That seems unlikely to be in line with what the user would want.

Relatedly, I've not had the time to try and understand it, but how does all this interact with courses? Are you supposed to set Stop Recording on music select, or result? If the latter, is it then possible to record a continuous recording for a course?

`MainController.java`
A question from a point of ignorance, should the `ObsWsClient`'s `.close()` method be called on game shutdown? Does it matter?



`ObsListener.java`
Minor question - was `triggerReplay` intended to be public?
Also, why does it schedule the second `triggerStateChange` to be executed later?


Lines 77 and 80:
Is the `isEmpty()` check redundant? It appears that way since it is immediately followed by an `equals` with a known constant.

The following code at line 83:
```java
if (instantStopRecord) {
	instantStopRecord = false;
	delay = 0;
}
```
I understand the intention is to immediately run the task that stops the recording, and this is entirely due my own lack of experience, but in order to convince myself that `delay = 0` will force `schedule()` to immediately invoke the lambda we pass to it I've had to journey through the oracle java docs, OpenJDK source, and step through what it does with a debugger, and I still couldn't find a definitive answer. To avoid this complexity, maybe we could directly do what `scheduledStopTask` does inside the aforementioned if block?


Lines 52-53:
```java
if (scheduledStopTask != null && !scheduledStopTask.isDone()) {
	scheduledStopTask.cancel(false);
	scheduledStopTask = null;
	return true;
}
```
This usage of `Future` is faulty and introduces at least one race condition.
Consider this valid execution ordering:

```java
[scheduler task - line 89] obsClient.requestStopRecord();
[main - line 52] if (scheduledStopTask != null && !scheduledStopTask.isDone()) {
[scheduler task - line 90] scheduledStopTask = null;
[main - line 53]	scheduledStopTask.cancel(false);
```

In such a situation, line 53 will immediately NPE.

This is trivially reproduced by inserting a `Thread.sleep(3000, 0);` both at the beginning of the task's lambda, and between lines 52 and 53 in `cancelScheduledStop`, and setting the recording delay to something short, such as 100ms. In practice this can reasonably happen if the scene changes right as the set delay ends.

Lines 119-122 and 538-541 in `ObsWsClient.java` appear to use a similar pattern but I didn't review whether they share this problem.

Another question that I have about this piece of code does `Future.cancel(boolean)` actually guarantee that the corresponding thread completes when we pass in `false`? Is it possible that it just started `requestStopRecord()`, `cancel()` returns before that finishes, then `triggerStateChange()` continues as if the previous scheduledStopTask was already done? I looked at the docs but again could not find a definitive answer.

NotAkitake:
> `Config.java:203-204`
> 
> ```java
> private Map<String, String> obsScenes = new HashMap<>();
> private Map<String, String> obsActions = new HashMap<>();
> ```
> 
> Because the new `Config` members are abstract `Map`s, this appears to insert a `"class": "java.util.HashMap"` entry in our `config_sys.json` file. Please keep in mind that those config files are intended to be user-readable and leaving in entries such as these is liable to create confusion at worst and looks bad at best. I understand the instinct to use `Map` member variables over the concrete `HashMap`, but in this case I'd suggest using the latter.

So.. by making it a HashMap<String, String> it would show in config without the "class" key? 
Edit: Can confirm it works, thanks.

> `MusicResult.java:329` I don't fully understand all of the checks that occur at lines 322-324, but at a glance, making `saveLastRecording` depend on them seems odd - for example, if I understand the admittedly-horrifying `BMSPlayer` constructor correctly, the following block of code won't be entered if the score from the current play won't be saved due to custom judge or another similar option. That seems unlikely to be in line with what the user would want.
> 
> Relatedly, I've not had the time to try and understand it, but how does all this interact with courses? Are you supposed to set Stop Recording on music select, or result? If the latter, is it then possible to record a continuous recording for a course?

If the player selected `On Replay` as his method to save recordings, then he should expect that to happen when a replay normally would save, which is in that function, in that if statement. So I completely disagree here.
As for the course, no it would start a new recording for each song currently. It's probably possible to make it record the whole course but I cba looking into it, if you want to give it a look go ahead but it'd be a huge headache, especially as the player is free to set whatever game state as stop/start record.

> `MainController.java` A question from a point of ignorance, should the `ObsWsClient`'s `.close()` method be called on game shutdown? Does it matter?

It doesn´t matter. The websocket connection will just timeout on the OBS side shortly after and it doesn't affect anything. It'd just be a redundant call.

> `ObsListener.java` Minor question - was `triggerReplay` intended to be public? Also, why does it schedule the second `triggerStateChange` to be executed later?

It's public because I used to call it from MainController, but you're right that it doesn´t technically need to be, although it doesn't technically hurt anything.
As for why the two calls to triggerStateChange, that's because when you quick retry, changeState doesn´t get called again since you're still in PLAY. So that function manually triggers what the MainStateListener would send if the player went back to MUSICSELECT, then a second later triggers PLAY again, so that if they set `Start Recording` on PLAY, it restarts the recording.

> Lines 77 and 80: Is the `isEmpty()` check redundant? It appears that way since it is immediately followed by an `equals` with a known constant.

You're right.

> The following code at line 83:
> 
> ```java
> if (instantStopRecord) {
> 	instantStopRecord = false;
> 	delay = 0;
> }
> ```
> 
> I understand the intention is to immediately run the task that stops the recording, and this is entirely due my own lack of experience, but in order to convince myself that `delay = 0` will force `schedule()` to immediately invoke the lambda we pass to it I've had to journey through the oracle java docs, OpenJDK source, and step through what it does with a debugger, and I still couldn't find a definitive answer. To avoid this complexity, maybe we could directly do what `scheduledStopTask` does inside the aforementioned if block?

Sure, though in my experience it's worked fine, if you feel it's safer to completely avoid the scheduler then I can change that.

> Lines 52-53:
> 
> ```java
> if (scheduledStopTask != null && !scheduledStopTask.isDone()) {
> 	scheduledStopTask.cancel(false);
> 	scheduledStopTask = null;
> 	return true;
> }
> ```
>
> This usage of `Future` is faulty and introduces at least one race condition. Consider this valid execution ordering:
> 
> ```java
> [scheduler task - line 89] obsClient.requestStopRecord();
> [main - line 52] if (scheduledStopTask != null && !scheduledStopTask.isDone()) {
> [scheduler task - line 90] scheduledStopTask = null;
> [main - line 53]	scheduledStopTask.cancel(false);
> ```
> 
> In such a situation, line 53 will immediately NPE.
> 
> This is trivially reproduced by inserting a `Thread.sleep(3000, 0);` both at the beginning of the task's lambda, and between lines 52 and 53 in `cancelScheduledStop`, and setting the recording delay to something short, such as 100ms. In practice this can reasonably happen if the scene changes right as the set delay ends.

Sure?

> Lines 119-122 and 538-541 in `ObsWsClient.java` appear to use a similar pattern but I didn't review whether they share this problem.

Well as a reviewer you're supposed to tell me, and again I'm new to Java so things like this aren't my forte.

> Another question that I have about this piece of code does `Future.cancel(boolean)` actually guarantee that the corresponding thread completes when we pass in `false`? Is it possible that it just started `requestStopRecord()`, `cancel()` returns before that finishes, then `triggerStateChange()` continues as if the previous scheduledStopTask was already done? I looked at the docs but again could not find a definitive answer.

I'll look into it.

core/src/bms/player/beatoraja/Config.java
```diff
@@ -192,6 +196,19 @@ public class Config implements Validatable {
 	 */
 	private String[] webhookUrl = new String[0];
 
+	/**
+	 * OBS WebSocket Control
+	 */
+	private boolean useObsWs = false;
+	private String obsWsHost = "localhost";
+	private int obsWsPort = 4455;
+	private boolean obsWsAuth = false;
```
seraxis: Does this config property need to exist? I think evaluating whether a password is empty or not in the javafx view and control class would be sufficient, though this is a minor issue.

seraxis (commented):
Leaving some feedback as a comment, I need more time for a more thorough review of the multi-threaded code. As far as the feature itself goes it's very impressive work! Great stuff

core/src/bms/player/beatoraja/MainState.java
```diff
@@ -190,5 +190,16 @@ public void stop(SystemSoundManager.SoundType sound) {
 	
 	public enum MainStateType {
 		MUSICSELECT,DECIDE,PLAY,RESULT,COURSERESULT,CONFIG,SKINCONFIG;
+
+		public static MainStateType fromString(String name) {
```
Catizard: Seemingly unnecessary because enum itself have `valueOf` function, also the code is dead

Catizard (approved):

PR #136, opened 2025-10-23
Improving config write safety and adding periodic saves

Arctice:
This PR improves beatoraja's handling of user configuration saving. First, it improves its reliability, going as far as to add a measure of crash-consistency, and adds periodic configuration file saves, addressing the outstanding issue of configs not being saved when the game is force killed, or the system crashes.

The original config write code did roughly the following:
```java
try (Writer writer = new OutputStreamWriter(new FileOutputStream(configpath))) {
    writer.write(toJson(config));
    writer.flush();
} catch (...)
```
This approach is very naive; it risks wiping the configuration file in the case of an error during json serialization, or an application or system crash; this is further exacerbated by the serialization taking up time after `FileOutputStream` was constructed, which immediately truncates the file. [0]

The first change in this PR introduces a `RobustFile` class that provides functions for writing files in a crash-consistent manner - even in the case of a system crash, either the write will complete succesfully or the original data will remain recoverable [1]. Files written using `RobustFile` must also be read using its `load` method [2]. I've include an explanation of the method used in the class file's comments. `Config` and `PlayerConfig` were changed to use the new functions.

The second change adds `MainController` code that periodically saves both the `Config` and `PlayerConfig` data to their backing files, currently set to roughly once every 5 minutes. This doesn't ensure that all changes made by the player will be immediately saved, but should prevent most cases of data loss while keeping the code relatively simple.  
(An approach where the config objects are marked on change and saved once soon after was tried, but this was more complex, and led to more widespread changes in the original code.)

With these improvements, future work on moving the launcher's configuration options into in-game menus in Endless Dream should be more tenable.

[0] The score database handling code internally relies on the C SQLite library, which already uses one of the most reliable file writing systems, and we essentially cannot improve on it.
[1] Hopefully - this stuff is difficult to get right and I'm not very experienced with it.
[2] `RobustFile` code is quite inefficient and shouldn't be used with large files or ones that require frequent writes.

### DIFF
--- a/core/src/bms/player/beatoraja/Config.java
+++ b/core/src/bms/player/beatoraja/Config.java
@@ -7,7 +7,9 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.text.ParseException;
+import java.util.logging.Logger;
 
 import bms.player.beatoraja.system.RobustFile;
 import bms.player.beatoraja.exceptions.PlayerConfigException;
@@ -16,6 +18,7 @@ import com.badlogic.gdx.math.MathUtils;
 import com.badlogic.gdx.utils.Json;
 import com.badlogic.gdx.utils.JsonWriter.OutputType;
 import com.badlogic.gdx.utils.SerializationException;
+
 
 /**
  * 各種設定項目。config.jsonで保持される
@@ -776,6 +779,7 @@ public class Config implements Validatable {
                 config = RobustFile.load(configpath, parser);
             }
             catch (IOException e) {
+                writeBackupConfigFile();
                 e.printStackTrace();
             }
         } else if(Files.exists(configpath_old)) {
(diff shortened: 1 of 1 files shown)


core/src/bms/player/beatoraja/PlayerConfig.java
```diff
@@ -1005,18 +1007,31 @@ public static PlayerConfig validatePlayerConfig(String playerid, PlayerConfig pl
...
+                                         0);
+            }
+        };
+
+        try {
+            PlayerConfig player = RobustFile.load(path, parser);
+            return player;
+        }
+        catch (IOException e) {
+            writeBackupConfigFile(playerpath, playerid, path);
```
Catizard: What's the reason to write backup when failed to read config file?

Catizard (commented):
Excellent work. Only one question below:

Arctice:
Still good for review/merging.

Ported the `writeBackupConfigFile` approach from PlayerConfig to Config - the configuration load will now create a secondary backup in case of a read failure, to avoid wiping out the entire configuration file that might otherwise be manually recoverable.

(In certain cases, the parser does manage to selectively clear only the fields with bad values, but sometimes the entire file gets defaulted out, and that's the case we need to prevent. There are also situations where the game refuses to launch because of a bad config but doesn't modify the file and I find this acceptable.)

PR #140, opened 2025-11-02
Patch out arbitrary IPFS limitations

pfych:
Not really priority or pressing, but it's always bugged me that there arbitrary limitations on where this feature can pull from.


seraxis (approved):
Makes sense I guess. I'm just questioning the future of IPFS in the project more broadly. I'll merge this for now but I have serious questions about the viability of using it in the future without concrete plans as to how you'd like to use it.

Merged

PR #142, opened 2025-11-05
Add a notification when next score won't be sent to IR service

Catizard:
This pr fixes a small issue that should be resolved in #122: notify user if next play won't be sent to IR.

### DIFF
--- a/core/src/bms/player/beatoraja/play/BMSPlayer.java
+++ b/core/src/bms/player/beatoraja/play/BMSPlayer.java
@@ -474,6 +474,9 @@ public class BMSPlayer extends MainState {
 		if (!score) {
 			ImGuiNotify.warning("Score nullifying options enabled. Next play will not be saved");
 		}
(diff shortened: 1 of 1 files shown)


core/src/bms/player/beatoraja/play/BMSPlayer.java
```diff
@@ -474,6 +474,9 @@ else if (RandomTrainer.isActive() && model.getMode() == Mode.BEAT_7K && RandomTr
 		if (!score) {
 			ImGuiNotify.warning("Score nullifying options enabled. Next play will not be saved");
 		}
+		if (forceNoIRSend) {
```
seraxis: Can this overlap with the previous !score notification?

Catizard: Currently no because freq isn't use `score` to nullify the record. But yes this is very unpleasant and should be revisited in the future

PR #148, opened 2025-11-09
In-game Profile Switcher

dgarciah98:
In-game profile switcher to not have to restart the game if you ever need to change profiles.

## What it does:
- Switch between any profiles inside the `player` folder through mod menu
- Load respective configurations, skins, scores, etc. by reloading every state and object that depends on `Config` and `PlayerConfig`
- Also loads IR connection if configured

## Demonstration:
https://github.com/user-attachments/assets/ccfe02a5-5a3b-420e-babd-facdfe4eeb2c

## Notes:
- It's made to only works when in music selector (`MusicSelector` state)
- Some refactoring was made in order to avoid duplicating code
- Tested with BokutachiIR v3.1.0, at some point it connected to the IR twice for some reason at startup but after updating my branch and the IR jar to 3.1.1 it didn't happen anymore
- I made sure that the skins, the scores and the IR connection (which I think it was the most important stuff) could change without breaking anything, but I think some more testing could be good since I could have perfectly missed anything

### DIFF
--- a/core/src/bms/player/beatoraja/MainController.java
+++ b/core/src/bms/player/beatoraja/MainController.java
@@ -364,7 +364,7 @@ public class MainController {
 		// Dispose MusicSelector to unallocate loaded skin
 		selector.dispose();
 		initializeStates();
-		initializeDependentMenus();
+		updateStateReferences();
 		triggerLnWarning();
 		setTargetList();
 
@@ -446,7 +446,7 @@ public class MainController {
 		}
 		loudnessAnalyzer = new BMSLoudnessAnalyzer(config);
     	initializeStates();
-		initializeDependentMenus();
+		updateStateReferences();
 		MiscSettingMenu.setMain(this);
 		if (bmsfile != null) {
 			if(resource.setBMSFile(bmsfile, auto)) {
@@ -570,7 +570,7 @@ public class MainController {
 		skinconfig = new SkinConfiguration(this, player);
 	}
 
-	private void initializeDependentMenus() {
+	private void updateStateReferences() {
 		SkinMenu.init(this, player);
 		SongManagerMenu.injectMusicSelector(selector);
 	}


core/src/bms/player/beatoraja/modmenu/ProfileSwitcherMenu.java
```diff
@@ -0,0 +1,62 @@
...
+    float relativeY = windowHeight * 0.04f;
+    ImGui.setNextWindowPos(relativeX, relativeY, ImGuiCond.FirstUseEver);
+
+    if (ImGui.begin("Profile Switcher", showPlayerManager, ImGuiWindowFlags.AlwaysAutoResize)) {
+      ImGui.text("Change profile while in-game");
+      ImGui.newLine();
+      ImGui.text("Player Profile");
+      ImGui.combo("##Player Profile", SELECTED_PLAYER, players, 4);
+      ImGui.sameLine();
+      boolean clicked = ImGui.button("Switch!");
```
Catizard:
```suggestion
boolean canClick = main.getCurrentState() instanceof MusicSelector
              && !config.getPlayername().equals(players[SELECTED_PLAYER.get()]);
ImGui.beginDisabled(!canClick);
boolean clicked = ImGui.button("Switch");
ImGui.endDisabled();
if (clicked) {
    // ......
}
```

Catizard (commented):
Great work. I took a skim through it and it's pretty clean and didn't find any obvious issues. Two suggestions and a question:

- When stoping user from switching, it would be better disabling the button directly rather than doing nothing
- Indent in this pr is misaligned, sometimes 2 spaces sometimes mixed
- What's the commented code for? One in `PortAudioDriver` and one in `SystemSoundManager`.

Need further discussion:

- The switch button could be merged into misc settings. Kinda weird it has its own menu.
- I'm a little worried about the implementation of `loadNewProfile` since althought it's extracted as a function from `MainController`'s initialize steps, but it's not being reused. For example, I notcied that `SongManagerMenu.injectMusicSelector` has been called both in `MainController::create` and `MainController::loadNewProfile`. This requires further devs to remember they need to update the selector reference in `loadNewProfile`. (This is clearly beyond this pr's purpose, it's mainly because the initialization steps are not split cleanly)
- Does `initializeStates()` causes memory leak? For example, it creates a new `MusicSelector` instance but if the old one still has some references it won't be cleaned. This concern maybe minor because it might need to switch the profile for many many times to finally cause affection.

Catizard (approved):
I think this pr it's already good to go. Although the profile switch button's position might be a little bit weird. Waiting sera's opinions.

Also don't forget to remove the unused code in final version :)

PR #152, opened 2025-11-19
Refactor logging mechanism into slf4j

Catizard:
> Although this pr is targeting `0.3.1` but it's an ongoing proposal that has no motivation to be merged into `0.3.1`

This is a proposal for adapting slf4j as the logging interface. Replacing the old string concat operations with the static strings. This can bring a little performance improvement.

> Why is concatenating string in logging message is bad?
> Because it would be executed no matter log level we currently at. Suppose you write a log message like `Logger.getGlobal().info("Error message:" + e.getMessage())` and the inner expression `"Error message:" + e.getMessage()` is evaluated even we're currently on `SEVERE` level. Using the parameterized api provided by slf4j can avoid this useless concatenation

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -170,6 +170,7 @@ dependencies {
     implementation(libs.ebur128java)
 
     implementation(libs.javawebsocket)
+    implementation(libs.bundles.slf4j)
 
     // non-gradle managed file dependencies. jportaudio not on maven. "custom" scares me.
     implementation(":jportaudio")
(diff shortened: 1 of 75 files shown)


seraxis:
I'm a fan of this, it's a lot of code to touch but I think the log situation currently is untenable and needs fixing.

Catizard:
There're 2 unmentioned thoughts, discussed internally:

- Currently we're using JUL backend, wondering if other backends can provide better performance?
- Logging to a xml file is a big waste. How about migrating to plain txt? 

I think these two are beyond this pr's purpose. so I think there should be nothing blocks this pr anymore

seraxis (approved):
Merged

Issue #83, opened 2025-09-23
BMS downloads in background cause stutters during gameplay


saintnoodle:
Pausing downloads during gameplay would most likely avoid this issue, if possible.

PR #154, opened 2025-11-20
Batch downloads based on difficult table

Catizard:
image

This pr adds a new context menu item for difficult table: `Fill Missing Charts`. When selected it would scan and submit all missing charts to the download task queue.

As #83 indicates, the in-game download feature **can** have a big performance impact. Based on the complexity/ux issue stated in #84, my alternative idea is giving user a configurable option: `Auto extract & update download directory`. By checking this option, in-game download feature would no longer start extracting and updating download directory automatically. Instead, it records the task needs to be executed and leaving user a button to manually trigger the process. Need further discussion.

Another improvement is currently we're not able to open context menu on a folder that is under a difficult table. Which means we can only fill missing charts based on difficult table
EDIT: We can batch download based on folder now

### DIFF
--- a/core/src/bms/player/beatoraja/MainController.java
+++ b/core/src/bms/player/beatoraja/MainController.java
@@ -1033,8 +1033,12 @@ public class MainController {
 	private UpdateThread updateSong;
 
 	public void updateSong(String path) {
+		updateSong(path, false);
+	}
+
+	public void updateSong(String path, boolean updateParentWhenMissing) {
 		if (updateSong == null || !updateSong.isAlive()) {
-			updateSong = new SongUpdateThread(path);
+			updateSong = new SongUpdateThread(path, updateParentWhenMissing);
 			updateSong.start();
 		} else {
 			logger.warn("楽曲更新中のため、更新要求は取り消されました");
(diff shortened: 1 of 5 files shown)


seraxis:
As discussed internally we should extend this PR's functionality to include:
- Handle httpdownloader being disabled
- Handle an existing HttpDownloadSource being disabled
- Handle downloading missing course songs
- Configure where http_download goes

This last one requires some discussion and forethought as http_download is currently special cased outside of the users `resources`

Catizard:
Custom download directory feature transferred to #156

Catizard:
We need to refactor the usage of `updateDirectory`. This function is called each time we download a package successfully. It has two possible ways to do this:

- Disable automatically calling this function and let user update the download directory manually(as stated in this pr's comment)
- Currently we're passing the parent directory(`http_download` as the default directory) to `updateDirectory`, it's the root cause that it's slow. A straightforward idea is replacing it with individual directory (example: `http_download/Freedom Dive`). However, this function's implementation doesn't support this. So we need to refactor the implementation to allow it update an inner directory, which might beyond this pr's purpose.

If we want to merge this before releasing 0.3.1. I would suggest to implement the first strategy as a temporary solution.

Catizard:
Converting this pr to draft again. Discussed internally we think we should implement the inner directory update strategy in this pr.

PR #156, opened 2025-11-21
Allow user customize their download directory

Catizard:
image

### DIFF
--- a/core/src/bms/player/beatoraja/launcher/ResourceConfigurationView.java
+++ b/core/src/bms/player/beatoraja/launcher/ResourceConfigurationView.java
@@ -60,7 +60,9 @@ public class ResourceConfigurationView implements Initializable {
 						super.updateItem(item, empty);
 						if (item != null && !empty) {
 							setText(item);
-							if (item.equals(downloadDirectory)) {
+							String entryAbsolutePath = Path.of(item).toAbsolutePath().toString();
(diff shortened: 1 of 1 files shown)


Catizard:
image

I uploaded a refactored version that changes the interface as asking user to mark one of the directories in resources tab as download directory instead of providing an input box.
I think the old version **can** serve well. The later refactored one is a little more complex but has a better looking.

Currently it has an unresolved issue: it cannot display the arrow down icon correctly(as the picture shows, the third button is a question mark)

seraxis (approved):
Clean implementation, I like the solution presented here. Minor comment about the button display but otherwise good to merge.

Catizard:
image

Remove the text icon and add a tooltip for this button.

There's a migration issue discussed internally: as shown in this picture, the `http_download` directory here in resources tab, it's an absolute path. However, the default value of download directory, it's `http_download`, which is a relative path. This causes the users migrated from previous versions, the download directory won't be highlighted (It actually doesn't affect the functionality). To fix this issue, 2ab67a7 changes the comparison between entires from resources tab and download directory from config file from plain string comparison to convert them as absolute path first. By doing this we can handle the compatibility from previous versions.

Also, this feature should be working fine for a new user, either creating his own download directory or not:

- if he chooses to not create a new directory and mark it as the download directory. Everything is working as the previous versions. A default directory `http_download` would be created after pressing the play button. And it would be highlighted correctly.
- if he chooses to make his own, httpdownloader would know and download to specified directory correctly.

PR #161, opened 2025-11-23
Fix the npe caused by SongDatabaseUpdateListener

Catizard:


seraxis (approved):
Thanks a bunch, merged

PR #164, opened 2025-11-24
fix for incorrect behaviour when searching with 0 results

Arctice:
The behaviour implemented in `3090e308` where opening the skin widget debugger (#79) blocks keyboard input created an issue where, when searching for something that gives no results, hitting return would incorrectly select the currently hovered folder or song. I fix.


seraxis (approved):
Merged, nice catch

PR #179, opened 2025-12-03
Goto result scene button

Catizard:
image

This pr adds a button that directly jumps to the result scene from music selector scene. For easier debugging and skin making. The loaded song/score would be the last chart player played. Things like replay and gauge data wouldn't work.

TODO: Some skins might cause immediate crash when pressing the goto button. This is because the implementation strategy is trying to mimic the normal game behavior. Demanding further tests.


seraxis:
This is cool, but I think this feature should be more fleshed out in a series of mock scenes that can be previewed quickly for skinning, instead of this kind of hack. I'm not sure I want to merge this feature as is

PR #186, opened 2025-12-07
In-game skin property documentation

Catizard:
> This pr's implementation is completed. It's lacking the documentation. So I'm calling helps for completing the skin property documentations.

image

This is a proposal for providing an in-game skin property documentation. As the picture shows, there's an expandable tree in game window and it shows the skin properties. If a skin property has javadoc in source code, it'll be rendered as well.

We have discussed internally about the documentation strategy. There's mainly two ideas:

- Introducing a new project that handles the documentation.
  - Pros:
    - Could be easier for contribution.
    - Can have better render strategy, like a website.
    - I18n support could be easier
  - Cons:
    - Introduces a new standalone project. So we need to put efforts in gluing two projects
    - Versions could be hard to keep the pace (i.e. if we change a skin property's behavior in ED 0.5.0, the documentation project needs to know that and mark the difference manually)
- Keeping the documentation inside same project
  - Pros:
    - Easier to keep the pace. For example, if a pr changes one skin property's behavior, we can easily check if this pr had changed the corresponding documentation or not.
    - No glue layer introduced. Implementation could be very easy.
  - Cons:
    - I18n support is almost impossible.
    - Need extra efforts to export the documentation outside.
    - Could cause many prs for changing the documentation. Put extra stress on maintaining the project
    - Hard for non-programmers to evolve. As the documentation is bound with source code

This pr picked the second way to have an easy and clean implementation for a start. If we have gathered many documentations in the future. It wouldn't be hard to migrate to the first way too.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -158,6 +158,10 @@ dependencies {
     implementation(libs.javawebsocket)
     implementation(libs.bundles.slf4j)
 
+    // runtime-javadoc is split into an annotation processor and a library
+    annotationProcessor(libs.runtime.javadoc.scribe)
+    implementation(libs.runtime.javadoc)
+
     // non-gradle managed file dependencies. jportaudio not on maven. "custom" scares me.
     implementation(":jportaudio")
     implementation(":luaj-jse:3.0.2-custom")
--- a/core/src/bms/player/beatoraja/MainController.java
+++ b/core/src/bms/player/beatoraja/MainController.java
@@ -422,6 +422,10 @@ public class MainController {
             ImGuiRenderer.init();
         }
 
+		try (var perf = PerformanceMetrics.get().Event("Skin manual init")) {
+			SkinPropertyManual.init();
+		}
+
         try (var perf = PerformanceMetrics.get().Event("System font load")) {
 			FreeTypeFontGenerator generator = new FreeTypeFontGenerator(Gdx.files.internal(config.getSystemfontpath()));
 			FreeTypeFontParameter parameter = new FreeTypeFontParameter();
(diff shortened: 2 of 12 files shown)


Catizard:
Contribution guide for non-programmers:

Say you want to add documentation for string properties, press [link](https://github.com/Catizard/lr2oraja-endlessdream/blob/feat/skin_property_doc/core/src/bms/player/beatoraja/skin/property/StringPropertyFactory.java#L59) here. And press the edit button here
image
This will make you enter the edit mode. You can now modify the files directly. Suppose I want to add documentation for `genre(13)`. Copying the documentation from other properties and changing the text content and it'll be like:
image
After editing, pressing the right top corner button "Commit Changes", it would pop up a form like:
image
Commit message is the message for this change. It doesn't matter so you can type anything. And you'll have to select the second option `Create a new branch...` here. Below input box will be automatically filled, it also doesn't matter. Finally, by pressing the button "Propose changes" your work would be submitted to me and I can check them.

Links:

- String: https://github.com/Catizard/lr2oraja-endlessdream/edit/feat/skin_property_doc/core/src/bms/player/beatoraja/skin/property/StringPropertyFactory.java#L59
- Boolean: https://github.com/Catizard/lr2oraja-endlessdream/blob/feat/skin_property_doc/core/src/bms/player/beatoraja/skin/property/BooleanPropertyFactory.java#L331
- Float: https://github.com/Catizard/lr2oraja-endlessdream/blob/feat/skin_property_doc/core/src/bms/player/beatoraja/skin/property/FloatPropertyFactory.java#L306
- Integer: https://github.com/Catizard/lr2oraja-endlessdream/blob/feat/skin_property_doc/core/src/bms/player/beatoraja/skin/property/IntegerPropertyFactory.java#L451
- Event: https://github.com/Catizard/lr2oraja-endlessdream/blob/feat/skin_property_doc/core/src/bms/player/beatoraja/skin/property/EventFactory.java#L73

PR #188, opened 2025-12-09
Send gauge history to IR service

Catizard:
This pr packages & sends the gauge history to IR service

### DIFF
--- a/core/src/bms/player/beatoraja/ir/IRGaugeHistory.java
+++ b/core/src/bms/player/beatoraja/ir/IRGaugeHistory.java
@@ -36,8 +36,9 @@ public record IRGaugeHistory(List<Float> easy, List<Float> groove, List<Float> h
 		if (floatArray == null || floatArray.items == null) {
 			return ret;
 		}
-		for (float item : floatArray.items) {
-			ret.add(item);
+		// Don't use for-range on floatArray.items, its capacity can be different with size!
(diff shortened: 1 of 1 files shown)


core/src/bms/player/beatoraja/ir/IRGaugeHistory.java
```diff
@@ -0,0 +1,48 @@
...
+		}
+		return new IRGaugeHistory(
+				gaugeLog.length > 1 ? convert(gaugeLog[1]) : emptyGaugeData(),
+				gaugeLog.length > 2 ? convert(gaugeLog[2]) : emptyGaugeData(),
+				gaugeLog.length > 3 ? convert(gaugeLog[3]) : emptyGaugeData(),
+				gaugeLog.length > 4 ? convert(gaugeLog[4]) : emptyGaugeData()
+		);
+	}
+
+	private static List<Float> convert(FloatArray floatArray) {
```
Catizard: image
The `floatArray` here, is an array of gauge data like `[0, 0, 5, 12, ...]`. Its length is a calculated value based on the chart's time length. However, there's a difference between the "chart's time length" and the "chart's final note time length". Beatoraja allows you to quit the play scene after hitting the final note, which causes a small gap shown in the picture.

Catizard (commented):
Waiting fix

PR #200, opened 2026-01-05
Migrate from jbms-parser to kbms-parser

Catizard:
This is another radical proposal: replace `jbms-parser` with `kbms-parser`. The only benefit gained from this migration is `kbms-parser` has an open-source license. The original `jbms-parser` is running enough fast that no optimization could be done by rewriting it.

This pr is still in early development stage because keeping compatibility/correctness between two projects is hard.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -31,7 +31,7 @@ version = libs.versions.beatoraja.get()
 
 sourceSets {
     main {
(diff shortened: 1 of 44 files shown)


seraxis:
Really impressive work. I think this will be an important foundation to build off of if we can get it fast enough

PR #206, opened 2026-01-19
Real in-game setting system

Catizard:
image

## Motivation

Beatoraja exposes its configurable options in its launcher, which is known to be hard to use & no explanation. And most of the tabs are disabled after launching the game. To allow user change options in-game, beatoraja created many specific events for skin makers to use. Most of the popular skins have features like a panel that arranges some options. However, what options are presented is now fully controlled by the skin makers: what if the option that you want isn't presented in the skin you're using, what if we add more and more options in ED, what if the skin I want to use doesn't have this feature at all?

Previously, we have an idea of bringing the `lift`, `hidden`, `lanecover` settings in-game. I assembled them in mod menu's `Misc Setting Menu`, which is simply adding them all and blindly stacking them together. This proposal could be viewed as the extension of that.

This proposal can also unlock some other ideas like:

- Pressing tab key to open a "song info" panel that allows you do some things for current selected song, rendering the leaderboard and manage favorited random permutations.
- Making all shortcuts in beatoraja are configurable

## What can be done in this system

- Almost all configurable options
- Osu-style key configuration
- Giving user a button to fire an issue that attached with system parameters, configurations.

## Implementation details

Currently, it **overrides** the F1 key to open the menu.

The code could be roughly split into 2 pieces:

- Widgets: render things by using Imgui and they can expose each others
- Windows: display the configurable options

For reviewer and further dev:

- `ImGuiKeyHelper`: getting the last pressed key and translating it into libgdx's keycode. Libgdx's input processor is frozen when `SettingsMenu` is at front.
- F1 key is handled manually in `poll` function, to allow user close the setting menu by pressing F1 again.
- Font size cannot be configured dynamically at runtime, there's a specific 24 pixel sized font called `font24`.


BKVad1m:
> > Replace F1 -> F7 for show and hide FPS because twitter (X) was useless
> 
> F7 is too hard to press

but F1 for in-game settings stays

seraxis:
If I am understanding this correctly this current PR proposes to add a toggle for FPS inside the new in-game settings menu, as opposed to rebinding it to yet another key (such as F7), is that correct?

core/src/bms/player/beatoraja/audio/BMSLoudnessAnalyzer.java
```diff
@@ -65,12 +65,10 @@ public BMSLoudnessAnalyzer(Config config) {
 		this.available = checkLibraryAvailable();
 		this.cacheDir = Paths.get("cache/normalize");
 
-		if (config.getAudioConfig().isNormalizeVolume()) {
```
seraxis: is removing this intentional or a regression?

PR #213, opened 2026-01-28
Make all shortcuts able to rebind/disable

Catizard:
> **Warning**
> Don't run this branch if you're not familiar with how beatoraja handles the config files. This branch is rather unstable and there's a bug that purges all your config files when something went wrong. And there's no guarantee that this branch can work. I have warned you.

> Side note for devs: My initial idea was implementing a more concise abstract layer that would be inserted between input layer and upper components layer (e.g. `MusicSelector` currently is using the low-level input methods that directly tells you which key is pressed directly). However, this is harder than I expected so the implementation currently is implementing a small wrap layer for input handling.

This pr is an extension that based on #206.

## Motivation

Disabling some shortcuts or rebinding them. For example, a problem that haunts me for ages it's I cannot disable some shortcuts in beatoraja, especially F8 and F9 (mark song/chart as favorite or hidden), which is totally a garbage design.

## Implementation

Currently, this pr has achieved allowing user rebind/disable the most of the shortcuts in music scene. All modifications are persisted on disk.

## Drop-in compatibility

No affection for new users but could have some struggle problems if we want to change how we store the shortcut modifications after merging this pr.


Catizard:
Update: the proposal's implementation has been done. Still waiting #206 to be merged

PR #218, opened 2026-02-11
Improve the lr2 skin compatibility

Catizard:
> I'll update this description when having updates. Please view the text before asking questions.

> The pre-release jar file can be grabbed at https://github.com/Catizard/lr2oraja-endlessdream/releases before this pr is merged into main line

This pr aims to improve the compatibility of the lr2 skins, which is as known as pretty poor in upstream. The main goal of this pr is:

- Loading popular skins like `Seraphic`, `KCOOL`, `WMIX`, etc. as native lr2
- Improve the skin widget debugger experience

Details:

- A lr2 skin can have no resolution settings. Therefore, this pr adds an option for every lr2 skins allowing users to modify the resolution they want.
- A lr2 skin can have multiple options that with same name. This is handled by forcing one option can be used only once.
- Some features are not implemented in beatoraja(e.g. extra mode, battle) and could lead to unexpected behaviors. For example, `Seraphic` shows a pentagram picture as part of the background, however, it's written like:
```
if (extra mode on) {
    bg1
} else (extra mode off) {
    bg2
} else {
    ???
}
```
Because extra mode is not a thing in beatoraja, this image would simply be ignored. The fixing way is adding some values manually. See `bms.player.beatoraja.skin.lr2.LR2SkinLoader#LR2SkinLoader` for details.
- Dst option 80 is true when lr2 is loading a bms file, which is also implemented in beatoraja. Apparently this option should have no meaningful value at scene that is not play scene. However, it turns out it's always true at select scene in lr2. Currently, this value is modified manually in this pr too to make specific skin `Seraphic` to work.

> I know this is dangerous :(

- Buttons that opening the panel are not working: this is because beatoraja doesn't give a proper click event for button types from 1 to 9. However, there's an extra problem when you want to open the panel by clicking mouse rather than holding start/select key: beatoraja would try to clear the panel every render frame when no start/select key is pressed. This is handled by introducing an extra state.
- If no font is defined, the text would be drawn by using the default font. However, due to a simple code error nothing because upstream doesn't setup the string property correctly.
- Unlike beatoraja skins, lr2 doesn't use an id field to connect `SRC_XX`, `DST_XX` commands, instead, it forces the skin makers to put the definitions sequentially. This leads to a problem that skin widget debugger cannot assign a meaningful name for the widgets. Apparently this makes the debugging process extremely hard. So in this pr the debugger is also improved for debugging the lr2 skins:
  - Image, Text, Buttons would be set a name that tells their parameters to help finding where they're defined
  - Missing definitions would be shown in debugger too. For example, beatoraja doesn't implement the `text st=63(1P Random)`, this will be listed in the debugger instead of letting the issue runs away.
  - Skins' custom options and its value would be shown in debugger because beatoraja is using an `IntIntMap` to store the mappings, which doesn't have a proper viewing in java debugger.
- Commands `SRC_NUMBER` could cause NPE and produce weird behavior because upstream misunderstands the parameters provided for this command. The most famous example is using `KCOOL` as play skin in beatoraja, no judge and combo would be displayed during the play. 
- Commands `SRC_TEXT` with st=1 shows nothing. This is because what beatoraja defines for st=1 differs from lr2. In beatoraja, it's the rival's name or empty string if no rival is chosen. In LR2, it's rival's name or target score if no rival is chosen. However, beatoraja introduces a custom definition: target(3) that implements the same behavior. So the modification here is simply hijacking the `SRC_TEXT` commands if its st=1.
- LR2 allows user change the bms file's metadata by using the `SRC_TEXT` commands from 20 to 30, which is useless and not supported in beatoraja. Therefore, they'll be simply converted to the uneditable version instead.
- Some lr2 skins are using dxa file for compressing resources. DXA is a very niche format that doesn't have any library supported so I manually ported one into java included in this pr. Currently the way we support loading dxa files is by inserting a guard that checks if the loading file isn't exist, if so, try looking its parent to see if there's a dxa file that has the same name and trying to extract the dxa file directly at the same directory.
- Some skin uses an unexpected combination of single dst & non-zero time definition with loop=0, which will never be drawn on the screen. In this pr I provided a specific hack for LR2 skins to avoid this issue.
- Fix the issue of non-zero y value from SRC_NOWCOMBO command would be miscalculated.
- If an animation combines negative time and loop == -1, raja would play it indefinitely because it assumes the minimum value of time is -1. This is fixed by changing the minimum value from -1 to -INF
- Some properties are ment to be some other's strict reverse version. E.g. dst_op 174 (no readme) should always be a reverse of dst_op 175 (has readme). However due to the beatoraja's design sometimes they're returning the same value which breaks the assumption. Currently, some of them are modified as a reference to their siblings to keep the rule (e.g. 174's value is based on the value produced by 175, rather than calculating a value itself). **There should be a list of them. If not, please remind me to attach them.**
- In lr2, a positive rotating degree means to rotating clock-wisely. However, in libgdx is the opposite. So for lr2 skins the angle would be transformed when loading.
- Some LR2 result skins are using 30 to mimic the difficult table's name, this is implemented in this pr by rerouting: when a SRC_TEXT has st=30 and current scene is not music select, it will be transferred as 1003
- #IF/#ELSE statements are erroneous combined with #INCLUDE, e.g. trying to switch the side config in Seraphic's result skin will never take effect, beatoraja will first load the `result_left.csv/result_right.csv` file correctly, but then load the `result_normal.csv` no matter what, which overrides everything just loaded. This is because `LR2SkinLoader::processLine` is sharing the same context: this when executing the command `#INCLUDE`. More specifically, there're two internal states: `ifs` and `skip` in `LR2SkinLoader`, which are used to indicate the current state of the `#IF/#ELSE` branches. However, `this` from `LR2SkinLoader` is passed down when executing the `#INCLUDE` command, if the file specified by `#INCLUDE` also contains `#IF/#ELSE` statements, the internal states would be changed, leading to unexpected control flows. This is fixed by removing the internal states from `LR2SkinLoader` and forcing the function `LR2SkinLoader` use an explicit context object and be sort of pure.
- Gauge chart color in result scene is adapting the definition from #SRC_GAUGECHART. And its background color is forced to be transparent to mimic the LR2 design.
- Number 174 is not implemented before. I filled it in this pr and it has a connected issue that if player is playing this chart for the very first time, then old combo will be set to -INF, which causes old combo count and diff combo count doesn't be displayed. I canceled this behavior in this pr, instead, old combo is always 0 when no old score exists.

## Known bugs

- Button that open/close lane cover, change screen mode doesn't work.
- When opening the panel by clicking the mouse, pressing play keys cannot tweak the options.
- Ghost/Scoregraph won't be supported because they don't exist in beatoraja
- LR2 seems only having one replay save slot, which makes replay option won't be supported.
- Difficulty, EX Level, Insane Level won't be displayed because I don't know what they're
- Custom search text font won't be supported
- Star match in the middle of the path is broken on windows and unsupported. E.g. /a/*/main.png
- Special value CONTINUE for #IMAGE is not supported
- #SRC_BAR_FLASH and #DST_BAR_FLASH is not a thing in beatoraja, won't be supported in this pr too.
- Timer 151, 152 is not functional in Beatoraja, it's basically not ever implemented. And probably won't be supported in this pr because it affects the raja skins if they're using the 151, 152 timer.
- Based on last entry, the `start` and `end` parameters in `SRC_GAUGECHART_1P` is not supported too because raja simply doesn't work like that. And I don't have an idea how to control it.
- Correct implementation of 130\~134, 176\~178 requires to have the ability to add new properties, which depends on other pr so it won't be implemented yet.
- The copy of wmix ships with a mpg video file called `off.mpg` that cannot be loaded on macos, could be a platform issue. Other functionalities are fine.
- The buttons both placed on panel and has overlay animations, the animation will be played no matter whether its corresponding panel is shown or not.
- The song select bars in wmix select skin will render nothing when returning from play scene. This will be considered as a wmix skin issue rather than this pr's fault. The manual fix is finding the `#DST_BAR_TITLE` line and change the numbers at the end from `1, 0, 0` to `0, 0, 0`.
- The vertical gauge bar is not implemented correctly
- DST_GROOVE_GAUGE's angle parameter is not working

This pr is also opening for extending the definitions and supporting special lr2 hacks like f/s, arena, etc.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -157,6 +157,7 @@ dependencies {
 
     implementation(libs.javawebsocket)
     implementation(libs.bundles.slf4j)
+    implementation(libs.reflections)
 
     // runtime-javadoc is split into an annotation processor and a library
     annotationProcessor(libs.runtime.javadoc.scribe)
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -158,6 +158,10 @@ dependencies {
     implementation(libs.javawebsocket)
     implementation(libs.bundles.slf4j)
 
+    // runtime-javadoc is split into an annotation processor and a library
+    annotationProcessor(libs.runtime.javadoc.scribe)
+    implementation(libs.runtime.javadoc)
+
     // non-gradle managed file dependencies. jportaudio not on maven. "custom" scares me.
     implementation(":jportaudio")
     implementation(":luaj-jse:3.0.2-custom")
--- a/core/src/bms/player/beatoraja/MainController.java
+++ b/core/src/bms/player/beatoraja/MainController.java
@@ -3,6 +3,9 @@ package bms.player.beatoraja;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.*;
+
+import bms.player.beatoraja.modmenu.skin.debugger.SkinDebugger;
+import bms.player.beatoraja.modmenu.skin.debugger.SkinWidgetManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -422,6 +425,10 @@ public class MainController {
             ImGuiRenderer.init();
         }
 
+		try (var perf = PerformanceMetrics.get().Event("Skin manual init")) {
+			SkinPropertyManual.init();
+		}
+
         try (var perf = PerformanceMetrics.get().Event("System font load")) {
 			FreeTypeFontGenerator generator = new FreeTypeFontGenerator(Gdx.files.internal(config.getSystemfontpath()));
 			FreeTypeFontParameter parameter = new FreeTypeFontParameter();
@@ -448,6 +455,7 @@ public class MainController {
     	initializeStates();
 		updateStateReferences();
 		MiscSettingMenu.setMain(this);
+		SkinDebugger.init(this);
 		if (bmsfile != null) {
 			if(resource.setBMSFile(bmsfile, auto)) {
 				changeState(MainStateType.PLAY);
--- a/core/src/bms/player/beatoraja/MainState.java
+++ b/core/src/bms/player/beatoraja/MainState.java
@@ -1,7 +1,8 @@
 package bms.player.beatoraja;
 
 import bms.player.beatoraja.SkinConfig.Offset;
-import bms.player.beatoraja.modmenu.SkinWidgetManager;
+import bms.player.beatoraja.modmenu.skin.debugger.SkinDebugger;
+import bms.player.beatoraja.modmenu.skin.debugger.SkinWidgetManager;
 import bms.player.beatoraja.skin.*;
 import bms.player.beatoraja.skin.SkinObject.SkinOffset;
 import bms.player.beatoraja.skin.property.EventFactory.EventType;
@@ -115,9 +116,10 @@ public abstract class MainState {
 		if (this.skin != null) {
 			this.skin.dispose();
 		}
+		beforeSetSkin();
 		this.skin = skin;
 		if (skin != null) {
-			SkinWidgetManager.changeSkin(skin);
+			afterSetSkin(skin);
 			for (IntMap.Entry<Offset> e : skin.getOffset().entries()) {
 				SkinOffset offset = main.getOffset(e.key);
 				if(offset == null || e.value == null) {
@@ -131,6 +133,14 @@ public abstract class MainState {
 				offset.a = e.value.a;
 			}
 		}
+	}
+
+	private void beforeSetSkin() {
+		SkinDebugger.listenBeforeSetSkin();
+	}
+
+	private void afterSetSkin(Skin skin) {
+		SkinDebugger.listenAfterSetSkin(skin);
 	}
 
 	public void loadSkin(SkinType skinType) {
--- a/core/src/bms/player/beatoraja/constants/BGAExpandOptions.java
+++ b/core/src/bms/player/beatoraja/constants/BGAExpandOptions.java
@@ -1,0 +1,27 @@
+package bms.player.beatoraja.constants;
+
+import java.util.Arrays;
+
+public enum BGAExpandOptions {
+	FULL("FULL"),
+	KEEP_ASPECT_RATIO("KEEP ASPECT RATIO"),
+	OFF("OFF");
+
+	private final String name;
+
+	BGAExpandOptions(String name) {
+		this.name = name;
+	}
+
+	public String getName() {
+		return name;
+	}
+
+	public static String[] names() {
+		return Arrays.stream(BGAExpandOptions.values()).map(BGAExpandOptions::getName).toArray(String[]::new);
+	}
+
+	public static String getName(int v) {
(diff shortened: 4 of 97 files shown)


Catizard:
> Essentially posting the same info here, as I did on discord.
> 
> Assuming new op/st/type/num codes can be implemented, there are some QoL features that could be added.
> 
> EXHC clear for dst_option, so we can show specified image if a player has reached an EXHC clear. Essentially just an extension to op codes 100 - 105.

As https://lr2skinhelp.vercel.app/dstop.html indicates, LV[F] uses 106 ~109 as the extensions. This can be achieved by manipulating the lr2 skin loading process. Will be updated later.

> 
> For num codes:
> 
> * TOTAL value, so it can be displayed. T/N could be calculated in .csv in the cell. (According to https://right-stick.sub.jp/lr2skinhelp.html calculations are possible within a cell for num codes.)

LV[F] uses 301 as the TOTAL

> * Individual note types (scr, bss, ln), so we could display the amount of those notes. Total notes is already assigned to num 74.

This one is a little tough because no extension of LR2 has assigned any number for this. Introducing a new number definition is having a posibility of conflicting upstream or other extensions that we don't want to see. Apparently we want to avoid the case that we have to ask the users or skin makers to update their num definition after one version bump. We have an ongoing plan of adding new properties dynamically (#138). So this will be recorded after #138 is being merged.

> For button types, maybe the favorite/hide chart and song should be made as separate buttons? The current button type 90 (off/favorite/ignore) does not work as expected when clicked, but does display if a song is favorited or "ignored".

Honestly, I don't want to touch the favorite/hidden feature because we have an ongoing feature that replaces this garbage design.

Dugatee:
Here are the issues I've noticed using an unmodified version of Seraphic on the current pre-release:
- All rotation is backwards (a consequence of y-origin I believe)
- Animations with negative times don't loop properly (seems like loop = -1 literally sets the time to -1)
- When the cursor disappears, it doesn't reappear except on the play screen, or if the F5 menu is open

Select:
- Opcodes 10 and 12 don't work, they should be true when a DP chart is selected or DP filter is applied
- 174 and 196 are false when something other than a chart is selected, but LR2 skins expect them to behave like !175 and !197

Result:
- Gauge charts ignore the source and use default Beatoraja colors
- Gauge chart background should be transparent but uses Beatoraja default instead
- Gauge chart draw animation and corresponding timers (150, 151) missing
- Some logic to make update information work properly is still missing, but I haven't yet figured it out fully

Missing Features:
- Song diff table in select
- Custom cursor image

Catizard:
> > Result:
> > 
> > * Gauge charts ignore the source and use default Beatoraja colors
> > * Gauge chart background should be transparent but uses Beatoraja default instead
> > * Gauge chart draw animation and corresponding timers (150, 151) missing
> 
> These won't be implemented in this pr because iirc gauge chart is simply unconfigurable in beatoraja. If there's a future pr that supports configurable gauge chart then it could be an extension based on that.

Correction: gauge chart colors are configurable in beatoraja. Now gauge chart in result scene would apply the color defined by #SRC_GAUGECHART_1P command and force the background color to be transparent to mimic the LR2 behavior.

PR #219, opened 2026-02-14
Revert "Copy highlighted menu text in music select"

corndogit:
This reverts commit ad4ff2465231f677c2112dd7c8ffa6c221683709.

Since we already have the functionality to copy song title in the context menu, and the issues with it conflicting with keybinds, and a new bug that seems to cause a crash on Linux.

image


BKVad1m:
I have a feedback this for conflicting of C keybinding makes jams and not binding on keyconfig

seraxis:
This is sad, I think redundant design is still worth keeping if it's intuitive, however since this breaks current users setup I'm going to revert. I hope we can see this re-implemented in future

PR #222, opened 2026-02-23
Rework RivalDataAccessor to allow updating rival scores during play

Chazoshtare:
This PR introduces a huge rework to the game's RivalDataAccessor. It simplifies the way rivals are handled, allows us to lower the strain on IRs, and adds some basic support for lazy-loading rival scores for IRs that don't support downloading all scores at once (e.g. Bokutachi).

**Changes:**
- Almost everything related to rival handling is now moved to `RivalDataAccessor`, with the goal of making it a single source of truth about them. Retrieving scores, selecting the rival - everything happens there.
- Added a way to update their scores in local DBs. Previously the assumption was that it happens only once while the game launches with `Import rival scores` flag checked.
- On two events - (1) when you finish a song and IR returns scores to get your placement in result screen, and (2) when you scroll to a song and IR returns scores to show ranking - those retrieved scores are passed to the accessor, it checks whether some of them belonged to your rivals, and then updates them accordingly. This means that even without explicitly importing rival scores we still lazy-load them from IR without any additonal requests.
- General style, formatting, unspaghettifying improvements.

**Known issues:**
Lazy-loading a score doesn't refresh the song bar yet, meaning one has to close and open a folder to refresh the state. I want to address it in a separate PR if possible, but it's a bit more involved than just calling `.updateBar()`.

There is a potential data race that can happen if:
1. We start the game with import on, importing threads are started.
2. We look at a chart to lazy-load scores.
3. At the same time, import threads finish.

Although what can happen is that we'll load a score, save it, then invalidate the cache, meaning valid score will be pulled from it eventually anyway. Or maybe that we'll try getting a score from a cache that's being invalidated and get an empty result, that will lated be updated to a valid one. In both cases we reach eventual consistency.
Worst thing that can happen, is if we pull a score from threads and lazy-loading, and both return different results because our rival just played the song at the same time. Not only it's extremely unlikely, but also will eventually be updated by one of the mechanisms.
This is still better than beatoraja's original behavior, where the cache will save a null score for a chart you're looking at, and when import threads finish, it will never invalidate it until you restart the game.

**Notes:**
I tested this only with BokutachiIR, which requires unique player names (like any sane IR should). I sure hope that none of the IRs that we support allow players to set identical names, as this is the only way to distinguish who the score belongs to during lazy-loading (ID is not in the class fields, maybe even not returned). This is only a bigger problem if you have 2 players with the same name added as your rivals.
To test with Bokutachi, you need a custom build of it that returns your rivals. I want to contribute that there too, but not before this change is introduced, as that would make raja fire unsupported import requests.


seraxis (approved):
High quality work, great stuff. @Catizard @Arctice any thoughts

Catizard (approved):
Good work. The comment below is a tolerable issue that is tolerable because its existence is not guarenteed even in upstream and nobody is using it

PR #232, opened 2026-02-28
Update build dependencies and ensure more Gradle cache uses

Chazoshtare:
This PR updates several dependencies related to the build process, and updates parts of it that were required to change in new versions.

Most importantly, `build_time` property is removed (as discussed internally), and `git_commit` property is generated through a provider, while also being marked as an input of our custom metadata task. It ensures that Gradle can reuse its cache whenever applicable, as it won't be invalidated by the time changing. This will speed up the build in some cases, if one runs on a state of the project that's already cached.

Version updates themselves save about 1 second on both my Windows and Linux setups, although that of course depends on the machine. It's not a lot, but at the very least it doesn't run longer.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -1,7 +1,12 @@
 plugins {
-    id("org.gradle.kotlin.kotlin-dsl") version "4.2.1"
+    id("org.gradle.kotlin.kotlin-dsl") version "6.5.3"
 }
 
+java {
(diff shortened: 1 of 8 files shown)


seraxis (approved):
Good stuff, with the wrapper update and kotlin dsl tested we should revisit #165 though I think it could use a review from you. I'm not sure if I like placing inline classes in the build file, it feels unidiomatic to me

PR #240, opened 2026-03-10
Fix sometimes the texture is mixing up pixmaps multiple times

### LINKED ISSUE #210: [Bug] Improper display of PNG in sequential image BGA (especially with alpha transparency)
When displaying PNGs with transparency in a BGA, the game may add a thin outline that shouldn't be there, or may ignore alpha transparency entirely. This does not happen for me in LR2 or mbmplay, which is where these BMS were originally tested.

I've uploaded examples in the screenshot. The bugged versions are on the left, whereas the accurate versions from mbmplay are on the right. (Note that I didn't get screenshots of the exact same frames for both instances, but hopefully it's possible to tell what the general issue is.)
Espada and Apples use alpha transparency, but dvvt does not and yet still has a display error.

Download links of the affected BMS from the examples, for easy testing:
Espada de los angeles: https://drive.google.com/file/d/1fzGcpiUlAunjAtjgOxzw9oOpun1VjieK/view
Apples Apples Apples Apples: https://mega.nz/file/j250kYhS#kSOWsk-yrIHfrZEUYmNEXCxBHlwkEq9RUIBd9RB6kWs
dvvt11atrovv1cl111cl11yc2000: https://9domu46i.com/yuruyuru/phase22.html

This issue is also present in vanilla beatoraja and has frustrated me for a long time, as people moving over from LR2 means that their game is displaying the inaccurate versions of my BGAs...

Image

Catizard:
Close #210 

7798ae79661fadaa317a4ef40b543c9b

The reason why the bga is broken is because the cache strategy beatoraja codes. Somehow, it mixes up multiple pixmaps into one texture, eventually causing the image that has transparent pixels drawn on screen looks funny.

In this pr, I removed the complex strategy at all, replacing it with a mindless strategy which forces to create a new texture each time if the cache is not perfectly matched. Personally I don't think the original complex strategy is *that beneficial* on lowering the bga loading process pressure. So I think this replacement is *good enough*.

This pr might need further test on whether it breaks other bms' bga.

### DIFF
--- a/core/src/bms/player/beatoraja/play/bga/BGImageProcessor.java
+++ b/core/src/bms/player/beatoraja/play/bga/BGImageProcessor.java
@@ -25,19 +25,28 @@ public class BGImageProcessor {
 	 */
 	private Pixmap[] bgamap = new Pixmap[1000];
 	/**
+	 *
 	 * BGイメージのキャッシュ
 	 */
-	private Texture[] bgacache;
-	/**
-	 * キャッシュされているBGイメージID
-	 */
-	private int[] bgacacheid;
+	private Map<Integer, Texture> bgacache;
 
 	private final PixmapResourcePool cache;
 
 	public BGImageProcessor(int size, int maxgen) {
-		bgacache = new Texture[size];
-		bgacacheid = new int[size];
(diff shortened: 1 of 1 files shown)


seraxis (approved):
Looks good, merged

PR #246, opened 2026-04-22
Add personal best pacemaker target

corndogit:
This MR adds a pacemaker target for your personal best score between IR and local scores, whichever is higher.

This is a test done by temporarily moving my score.db to simulate a score existing locally but not on IR
Screenshot_20260422_151031

| Case | IR score | Target value |
| --- | --- | --- |
| Local score is higher than IR score | Screenshot_20260422_164724 | Screenshot_20260422_164816 |
| IR score is higher than local score | Screenshot_20260422_164205 | Screenshot_20260422_161225 | 
| Unplayed on both | N/A | image |

Limitations: ghost data is not used due to how the pacemaker targets work - instead it works in a similar way to the score grade targets

### DIFF
--- a/core/src/bms/player/beatoraja/PlayerConfig.java
+++ b/core/src/bms/player/beatoraja/PlayerConfig.java
@@ -71,11 +71,17 @@ public final class PlayerConfig {
 	 */
 	private String targetid = "MAX";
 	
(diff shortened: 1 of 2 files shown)


seraxis:
how does this impact existing skins? I'm worried about a potential index error situation but i guess the new string properties should make that not possible?

corndogit:
> how does this impact existing skins? I'm worried about a potential index error situation but i guess the new string properties should make that not possible?

Shouldn't impact skins - the options are populated from the list. Tested with a few skins that allow selecting pacemakers.

Only time I noticed a bug was when switching back from my JAR compiled on this branch to one compiled from develop, which resulted in the list showing as empty until it was scrolled a bit. This is an extremely small edge case though and would only occur if someone downgraded JARs while having personal best selected

image
image

PR #247, opened 2026-04-24
Add toggle for notifications in misc settings

MellDa1024:
image
This pr adds an option to toggle toast notifications.
When disabled, notifications are suppressed in the UI and redirected to the logs instead.

I originally planned to only disable screenshot notifications, but the code was getting a bit messy, so I decided to make it a global setting for all notifications.
I'm not sure if there are any critical notifications that should never be hidden, because this toggle will hide everything.

### DIFF
--- a/core/src/bms/player/beatoraja/modmenu/ImGuiNotify.java
+++ b/core/src/bms/player/beatoraja/modmenu/ImGuiNotify.java
@@ -3,6 +3,8 @@ package bms.player.beatoraja.modmenu;
 import imgui.ImGui;
 import imgui.flag.ImGuiWindowFlags;
 import javafx.util.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -11,6 +13,7 @@ import static bms.player.beatoraja.modmenu.ImGuiRenderer.windowHeight;
 import static bms.player.beatoraja.modmenu.ImGuiRenderer.windowWidth;
 
 public class ImGuiNotify {
+    private static final Logger logger = LoggerFactory.getLogger(ImGuiNotify.class);
 
     public static final float NOTIFY_PADDING_X = 20.0f;
     public static final float NOTIFY_PADDING_Y = 20.0f;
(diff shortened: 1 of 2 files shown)


Catizard:
I believe that the few error-level notifications in the project should always be displayed, regardless of the circumstances. For example the notification about the invalid play (some of them are 'error' level but some others are warning iirc)

MellDa1024:
If warning & error level should be always displayed, would it be better to rename the option and only disable levels below those? (None, Success, Info)

corndogit:
Perhaps an alternative would be the ability to set a minimum threshold for log messages. For example, setting it from a "minimum notification urgency" dropdown (I'm bad at naming things) for info, warning, error, etc. For better UX maybe show a tooltip/message if user has set the level to ignore everything to warn them that they could miss something important.

Since it looks like notifications that don't show are instead logged to the console output, and given that a reasonable default value would show all toasts, I think there is less to worry about if the concern is that a user will miss an error message.

seraxis:
I agree with corndog completely, if you can refactor this to display minimum notification urgency this would be a great addition

Arctice:
For the record, my idea for notification control was to allow the user to mute particular types of notification, so you could configure them individually to, for example, not show the screenshot notification. I'm not sure if adding a global switch like this is a great approach

(But adding this as a temporary solution seems fine by me)

MellDa1024:
There are a few minor things to consider regarding the implementation:

First, endlessdream's toast levels include SUCCESS and NONE, in addition to INFO, WARNING, and ERROR.
NONE seems to be a temporary level used when no type is specified, so it shouldn't be an issue (since it technically shouldn't occur in practice). However, I'm not entirely sure where SUCCESS should fit in the hierarchy—whether it should be treated as INFO level or prioritized even higher than ERROR. I tried to look for existing examples of SUCCESS toasts to get a better idea, but it seems it hasn't been used anywhere yet.

Also, I think further discussion is needed on how the suppressed notifications should be logged. Should they all be redirected to the logger as INFO, or should they follow their original toast level? (If the latter, handling SUCCESS will also need clarification)

Arctice:
I was planning on making it so all notifications always write to the log. Don't worry about that part

PR #251, opened 2026-06-03
fix for MPEG bga's starting at an unintended delay

Arctice:
raja starts BGA video playback at play ready time, then once the bms starts, seeks to the beginning of the movie to synchronize. (using `MovieSeekThread` in `FFmpegProcessor.java`)

this strategy has been unreliable for MPEG streams with `javacv` 1.5.9, as the `setTimestamp` method of `FFmpegFrameGrabber` would in some cases seek to a time around 1 second into the video, instead of the start, causing desynced playback.

this affected eg. the original `Love & Justice`, and `先天性エゴサ依存症候群`.

older versions of raja used `grabber.restart()` to restart the playback instead of seeking, but this seems to have been replaced with `setVideoFrameNumber` for performance reasons on looping BGA, which I didn't want to abandon without better understanding of the difference.

instead, I bump `javacv` to 1.5.11, which is more reliable at seeking through MPEG videos, and includes a fix to a memory leak in `FFmpegFrameGrabber`.


seraxis (approved):
Thanks

PR #254, opened 2026-06-08
Add command-based OBS WebSocket configuration

dj-kata:
This PR improves the OBS WebSocket integration in Endless Dream. The previous configuration UI was difficult to use because each screen had fixed scene/action selectors, so this changes it to a command-based setup.

## Changes

- Replaced the per-screen selector UI with a command registration UI
- Added a table for viewing and removing registered OBS commands
- Added `MAIN_START` and `MAIN_END` timing options for when the main game window starts and closes
- Added new OBS actions:
  - Show source
  - Hide source
  - Switch scene
- Kept the existing recording start/stop actions

The existing recording-related behavior is still available, while the new command list allows multiple OBS actions to be configured more flexibly.

image


seraxis:
I wonder if this is a good fit for the game in the first place. The original OBS configuration was intended to make saving replays easier and assist in some streaming automation. This feature is quite complex and maintaining it may be difficult in the future. 

I'd like to discuss this more before we think about merging this feature

dj-kata:
That’s a fair question, since I didn’t understand your intention.

I added this feature with the intention of making it a bit easier to use for streaming when integrated with OBS. I also think the ability to show or hide sources is a nice feature for streaming.
As for the recording feature, I incorporated it into the command UI because I thought it would be better to allow users to specify that the recording should cover everything from the start of the gameplay screen to the end of the results screen. However, since some OBS users prefer to use OBS’s own replay buffer feature to save only the gameplay they want, I’m not sure how much demand there actually is for this.

With the tool I’m developing, called [oraja_helper](https://github.com/dj-kata/oraja_helper), I provide a feature that enables automatic OBS control by registering fixed regions based on captures of each scene’s skin. I believe there would be significant benefits to integrating this into LR2oraja, as it would make the process much simpler.

However, the point that it would increase complexity is valid, so I think you should decide whether to implement this feature based on the anticipated demand.

PR #255, opened 2026-06-18
Add controller hotplugging

seraxis:
By extending ControllerAdapter BMSPlayerInputProcessor can directly listen to controller connect/disconnect events, we use this to allow for controller hotplugging that applies to all directinput devices, keyboards, and midi controllers

Currently tested only on MacOS ARM with USBemani and DCFish boards


seraxis:
Reviewed externally, merging

PR #259, opened 2026-06-22
Auto change download source to default if the using one is not existed

Catizard:
This pr auto switches the download source to default configured one if the download source configured in config file is no longer existed. Specifically wriggle.


Arctice:
While the wriggle downloader is obsolete, the code implementing it allows for substitute override-URL services with a very simple API. For example, I am currently building a server that makes use of this. We should discuss keeping this functionality in some form, as changing this part of code every time a new service is created or dies is stupid.

PR #272, opened 2026-09-05
fix: .fnt bitmap font text pixelated on strict GLSL compilers

julianZ99:
### Problem
Bitmap-font (.fnt) text renders pixelated on strict GLSL compilers.
Reproduced on Linux + Mesa (AMD RX 7800 XT).
Skins: ECFN select, Starseeker play/result.

### Root cause
glsl/bilinear.frag uses textureSize() (GLSL 1.30+) with no `#version`,
so strict drivers treat it as GLSL 1.20 and reject it. ShaderManager
returns null, the renderer falls back to the default SpriteBatch shader,
and since TYPE_BILINEAR isn't in setFilter, the sheet is sampled with nearest so it shows pixelated glyphs in game.
This problem is driver strictness dependant.

### Fix
Make the existing bilinear shader GLSL 1.20 compliant and feed it the one
value it can no longer query itself.

### Before / after
image
before

image
after

image
before

image
after

### Testing done
- Linux/Mesa: ECFN select + Starseeker play/result now smooth
- Default JSON (TTF), .sdf fonts, video/layer shaders unchanged


Arctice:
Awesome, looks good. This might just need a check to make sure nothing's broken on Mac, I think it was fussy about the shaders before

radiden:
lol i patched basically the same thing a few minutes ago, you should also patch ShaderManager to add logging so it doesn't error out silently

```diff
diff --git a/core/src/bms/player/beatoraja/ShaderManager.java b/core/src/bms/player/beatoraja/ShaderManager.java
index ef0a294d..fd79b491 100644
--- a/core/src/bms/player/beatoraja/ShaderManager.java
+++ b/core/src/bms/player/beatoraja/ShaderManager.java
@@ -5,18 +5,23 @@ import java.util.Map.Entry;
 
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.graphics.glutils.ShaderProgram;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ShaderManager {
-
+	private static final Logger logger = LoggerFactory.getLogger(ShaderManager.class);
 	private static HashMap<String, ShaderProgram> shaders = new HashMap();
 
 	public static ShaderProgram getShader(String name) {
 		if (!shaders.containsKey(name)) {
 			ShaderProgram shader = new ShaderProgram(Gdx.files.classpath("glsl/" + name + ".vert"),
 					Gdx.files.classpath("glsl/" + name + ".frag"));
-			if(shader.isCompiled()) {
+			if (shader.isCompiled()) {
 				shaders.put(name, shader);
 				return shader;				
+			} else {
+				logger.error(shader.getLog());
+				return null;
 			}
 		}
 		return shaders.get(name);
```

julianZ99:
Great minds think alike lol.
Nice idea, just pushed a commit adding the error log.

Catizard:
Can confirm this pr itself doesn't affect macos as when running on macos the shader files won't be compiled before this pr. A related pr is #226, which is trying to make the shader files to be compiled on macos. I think these 2 pr have overlapped in some bits but i'm not sure because #226 is being written months ago.

julianZ99:
Thanks for the info. I looked at #226 and at #240, which later fixed the transparency and blackframe problems reported in #210 by changing the texture cache.
there is some overlap in the bilinear shader on Linux. This PR keeps it compatible with GLSL 1.20, while #226 loads GLSL 150 variants on both macOS and Linux. Since the app only explicitly requests an OpenGL 3.2 context on macOS, I’m thinking whether the GLSL 150 variants are necessary on Linux as well.

Would it make sense to use the GLSL 150 variants on macOS, while Linux and Windows use the GLSL 1.20 shader from this PR?

If both changes are combined, `bilinear_unix.frag` would also need to declare and use `u_textureSize`; otherwise, the setter added by this PR would fail at runtime.

Is #226 still expected to be merged as-is? I can update this PR if any changes are needed once that’s decided.

PR #273, opened 2026-09-13
fix: prevent memory allocation on each keypress

zkldi:
raja pre-allocates `10000` KeyInputLog items before gameplay starts on each chart, so that no memory allocations happen for the first 10,000 keypresses of the chart.

However, the index isn't reset to 0 between charts, **so after a player has made 10,000 keypresses across their session, every new keypress becomes a memory allocation.** This is likely behind some serious in-game stuttering.


Catizard (approved):
Looks sane to me.

When looking at the patch initially I'm pretty worried that the old system is running on this giant error for years and nobody found it, which is normally a sign of "it's a bug but we rely on it" moment but after thinking for a while it looks fine to me

PR #274, opened 2026-09-13
fix(security): prevent in-game-downloading from overwriting any file on your pc

zkldi:
at the moment, files downloaded from the in-game downloader aren't checked to have safe paths. a downloaded zip that looks like `../foo.txt` or `/etc/passwd` will attempt to write files there on your PC. A malicious BMS server could unexpectedly do arbitrary things to files on your computer.


Catizard (approved):
Pretty clean

PR #275, opened 2026-09-16
feat: backbeat support

zkldi:
Adds support for [backbeat](https://backbeat.ac) to endlessdream. This doesn't make it mandatory or anything. Users that have no idea what backbeat is are completely unaffected by this, and won't even know it's there.

**DISCLOSURE:** Some code here was written with the assistance of AI - namely GPT-5.6 Sol, especially around the gradle stuff and java SDK including, which I don't really understand at all. I am _not_ an expert java programmer by _any_ means and so I used it as a heavy reference for implementation here. I **HAVE** read over every single line in this PR, and the non-sludgey-evil-java-build-tool-shit was written by me. The entire design of backbeat has Zero AI involvement - I am not a sludgeman, slopping on your repo.

:warning: This PR is blocked on https://github.com/seraxis/jbms-parser/pull/5

# What is Backbeat?

Backbeat is a library that defines a standard for a sort of "special" songs folder on your PC. It does fancy CAFS/sqlite stuff to get high performance, and uses package-manager-like checksumming to deduplicate assets. I've written a lot more on it here: https://backbeat.ac/docs

# How have you integrated it?

There's a [Java SDK](https://github.com/zkldi/backbeat) that is shipped here and linked to. We call methods from that in order to do backbeat stuff. If the user doesn't use backbeat, nothing happens. If the user does, charts, tables, courses and packs they have installed Just Appear in the game.

## Resource
Firstly, I've added a struct called `Resource` that is passed around instead of Paths. One requirement of Backbeat is that sometimes, small files are just served in-memory instead of being paths on the filesystem. This means that a couple places in the code (asset loaders) need to take `Resource` instead of just `Path`. This works out fine, in rust speak:
```rs
enum Resource {
  // tiny ones (<16KiB) are just in the sqlite db, so they don't have a path
  InMemory(Vec<u8>),
  // large assets are on disk
  Path(Path),
}
```

Then some methods on that to get a cursor/the bytes.

The small-file optimisation here has massive performance improvements for BMS - this is roughly what you guys were simulating with the `.7z` stuff, only in a way where you can amend sabuns/don't have to do decompression logic/can index files.

## Backbeat Bundles

It's worth looking at a [Backbeat bundle](https://data.makiba.ac/bundles/b-e4973dedfae6dcf43f24bc41559298fbccc84bafccf3f95fff7e2a311317d20b):
```json
{
  "filename": "ERIS[MX].bms",
  "assets": {
    "#BANNER 01.png": "d4dad70f8153f5ce7ebbdd2e2dd62d0289c37ace0e56417d82f10945ff4e98cf",
    "#BANNER 02.png": "eb5b21399258590335c703fe4f6e064eb3c5c74ff47368da41eb57d24278ac3e",
    "#BANNER 03.png": "cc51aef0dd3db9e5fe3948972b23e2f38188b6d31bd6954374c817c40a9e7103",
    "#BANNER 04.png": "d3ccb61b02cdc6c49e11347202fb711fb0dcc56d4056f4076f3e351a967d4b60",
    "#BANNER 05.png": "f3a651fcb6bc839d98fa1204dd72febc229100f388d5060f1582474109060ea5",
    "bass1.ogg": "bcfe2d82cf83d17b5366c4838fdad05a8d0bd64ad65829f777b2db6e43d90a5b",
    "bass2.ogg": "8ca231750dfb77857bd5ceb0036ee82d0ba5723068e3cf6d5a64114c1977b615"
// TRUNCATED
  },
  "desc": "Grand Thaw - ERIS MX",
  "chart": "H4sIAAAAAAAC..... TRUNCATED" // n.b. the original chart data is base64(gzip)'d. 
}
```
The TL;DR is that you get a chart file, then you strap on-to that an `assets` map. That asset map says:
> When this chart asks for `bass1.ogg`, give it a file with sha256 = `bcfe2d82cf83d17b5366c4838fdad05a8d0bd64ad65829f777b2db6e43d90a5b`.

If you've worked on distributedy systems or know a bit about nix/git/pnpm before you can probably fill in the blanks here - this lets us have distributed chart asset servers (because of the checksums), and dedupe shared assets.

## `ResolvePath(string) -> Resource`

Given that, I've implemented a new function on `SongData` - `ResolvePath` which takes... a path. and returns... a Resource.

We add one column to the database, `backbeat_bundle_id`, which is set if the chart was pulled from the backbeat store. Then, the implementation of `ResolvePath` ends up looking like:

```rs
if self.backbeat_bundle_id {
  return bkb_resolve_path(self.backbeat_bundle_id, path);
} else {
  return Resource::Path(self.folder + path)
}
```

Most of the codebase has been re-jigged around to pass around `SongData` instead of a path to a bms file - as with backbeat, you no longer necessarily have a path to the bms file - it might just be in the store.

There's also a method `ResolvePathWithExtensions` which tries a bunch of extensions - the same as usual. There's quite a bit of code that gets removed from raja with this method, as this logic was duplicated in a couple places before.

## Pulling Charts

We pull charts as part of the db loading, like it's just another source for charts. There's a method `pullFromBackbeat` which essentially does:

```rs
for chart in store.list_charts("bms", "bml", "bme", "pms", "bmson") {
  do_normal_stuff_with(chart)
}

# At a high level

Users can now download content with backbeat and it Just Appears in their game. Also, there's an implementation for packs, courses and tables.

Obviously this is a fairly large PR for fairly weird software with no real rhythm game precedent, but it's purely opt in.

# Testing

It's been tested by me for the past couple months, and about 10 other people for the past couple days. It's been tested on linux/macos/windows and seemingly has no issues. A couple people have had complete sessions with backbeat downloaded content. Performance of backbeat-content is significantly better.

Message me on discord with any questions (@zkldi.dev). Apologies for the large PR.

### CODE REVIEW

File: core/src/bms/player/beatoraja/PlayerResource.java [hunk 1]
```diff
@@ -87,16 +86,17 @@ public final class PlayerResource {
...
+	private SongData[] autoPlaySongs;
 	private boolean loop;
 	
 	/**
 	 * コース
 	 */
 	private CourseData coursedata;
 	/**
-	 * コースのBMSモデル
+	 * 読み込み済みのコース
```
zkldi, line 97: i don't speak japanese, this comment update was helpfully suggested by a robot. 🤖 🙃

